PR #4108, opened 2015-01-07
change the default spec argument

ticoann:
need to update https://github.com/dmwm/WMCore/pull/5555 together

### DIFF
--- a/src/python/T0/WMSpec/StdSpecs/Express.py
+++ b/src/python/T0/WMSpec/StdSpecs/Express.py
@@ -465,46 +465,37 @@ class ExpressWorkloadFactory(StdBase):
 
     @staticmethod
     def getWorkloadArguments():
+        """
+         If arg is not specifyed, automatically set by following default value
+        - default: None
+        - type: str
+        - optional: True
+        - assign_optional: True
+        - validate: None
+        - attr: change first letter to lower case
+        - null: False
+
+        """
         baseArgs = StdBase.getWorkloadArguments()
 ##         specArgs = {"Outputs" : {"default" : {}, "type" : dict,
 ##                                  "optional" : False, "validate" : None,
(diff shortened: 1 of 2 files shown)


ticoann:
@hufnagel, please review.

hufnagel:
Looks good, I assume I can merge this as soon as you merge the relevant WMCore code ?

hufnagel:
What about the PromptReco spec though ? That's in WMCore and can in principle be used by both normal central Ops requests (if they so desire) and the Tier0. So I need to override a few more things there, as before ?

ticoann:
I clean up a little further. Please review. For the PromptReco, I am not quite sure how to set the flag as you mention when we chat. For now I update necessary values (CouchCacheID, etc) as optional arguments
https://github.com/dmwm/WMCore/pull/5555/files#diff-90f4b259876ad7414f235789db2dbae4R366

hufnagel:
Still don't really like it, especially in light of what we discussed about the PromptReco spec.

Can't we get rid of the "useReqMgr = False" statements and the Repack and Express specs just override the appropriate parameters to make them optional ? I think that's cleaner. PromptReco could do the same (and then stays usable for both Tier0 and Requests).

ticoann:
I think you are right that is cleaner I will do that way. I was just trying remove duplicate definition if it is commonly used, but I agree that it is not good to have flag to do different things in the same method

ticoann:
Hi Dirk, I updated code as you suggested. dmwm/WMCore#5555 is updated accordingly

hufnagel:
Looks good, but let me run a Tier0 replay to check it out...

hufnagel:
Replay bombs, commented on the reason in #5555

hufnagel:
With the changes in dmwm/WMCore#5555 everything works now.

PR #4111, opened 2015-01-29
Update tier0 sls alarms scripts

lucacopa:
Also updates an old query in the diagnoseActiveRuns script

### DIFF
--- a/bin/cmst0_backlog_wma
+++ b/bin/cmst0_backlog_wma
@@ -1,100 +1,212 @@
 #!/usr/bin/env python
 
-from WMCore.Configuration import loadConfigurationFile
-from WMCore.Database.CMSCouch import Database
-from string import Template
+"""
+_cmst0_backlog_wma_
+
+Loof for created jobs by Workflow type in the Tier-0
+If the creted jobs overall by workflow type exceeds a configured threshold
+then it alarms about it.
+
+Availability metrics are defined as:
+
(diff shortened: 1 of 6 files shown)


lucacopa:
We are working on the other scripts, so far cmst0_late_workflows is working.
Do we need to change something in the xml? I am not sure this line is still valid:
<serviceupdate xmlns="http://sls.cern.ch/SLS/XML/update">

lucacopa:
@hufnagel please have a look to the cmst0_backlog_wma script. It is using the T0AST instead of couch db jobs. I tested this script too, adn It is working. I am running all the sls alarm scripts in a cronjob in vocms015:

#SLS Alarms for the Tier-0 - luis89@fnal.gov
*/10 * * * * /data/tier0/sls/scripts/runSLSAlarms.sh > /dev/null

The xml files are created each 10 min at /data/tier0/sls

hufnagel:
New cmst0_backlog_wma script looks good. Do you want to run these scripts agains various replay and the prod instance for a while to test them ? Or are you confident they do what they claim they do ?

If the latter, I could merge and we could start talking to Diego about running it on his monitoring node.

lucacopa:
I can run them with the prod instance for a while. This will let us tune the SLSConfig again (Right now the thresholds are just like before). I let John test them with some replays too.

PR #4115, opened 2015-03-14
Add hcalnzsRun2 Scenario

lucacopa:


lucacopa:
@hufnagel this patch is tested

PR #4124, opened 2015-04-29
Do except as -- the modern way

ericvaandering:
I don't know Dirk, this looks pretty risky. :-)


hufnagel:
Haha, ok, so I was using the correct syntax in most places already :-)

PR #4139, opened 2015-06-04
Configuration updates for Run2015A

lucacopa:
Configuration file has been cleaned up so we only have the PDs that we are using or
we have used during Commissioning2015.

Also, I added some release overriding for single step Express jobs.


lucacopa:
@hufnagel please review

hufnagel:
A few issues:

- hltmon commented out area should be removed
- actually not sure about the MinimumBias dqm_sequences, asking for feedback on this
- the comments still talk about 2012, we should make one section for 2015
  (with the eventual idea to also change the replay config to include that and one extra legacy section)

lucacopa:
I removed the hltmon commented out area and also updated the comments.

PR #4141, opened 2015-06-04
Fixed query for getting the HLT config

lucacopa:


lucacopa:
This is working fine in the prod tier0

PR #4145, opened 2015-06-10
Add new Replay Configuration for Run1 scale tests

lucacopa:
- Added ReplayOfflineConfigurationRun1: No alca producers or specific dqm sequences.
- ReplayOfflineConfiguration.py was updated according the current produciton tier0 config.
- Fix space in ProdOfflineConfiguration

### DIFF
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -61,7 +61,7 @@ setPromptCalibrationConfig(tier0Config,
 
 
 # Defaults for CMSSW version
-defaultCMSSWVersion = "CMSSW_7_4_4_patch2"
+defaultCMSSWVersion = "CMSSW_7_4_5"
 
 # Configure ScramArch
 setDefaultScramArch(tier0Config, "slc6_amd64_gcc491")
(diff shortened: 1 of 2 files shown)


lucacopa:
@hufnagel please don't merge yet, I have to cross check with DAS too

hufnagel:
A few comments:
 - Run1 replay config does not need Run2 scenarios or Run2 repack/express overrides
 - Run2 replay config does not need Run1 scenarios or Run1 repack/express overrides
 - Run2 prod and replay config should remove dqm_sequences from everything but SingleMu
      (SingleMu is the only one we got a confirmation for what we use)
 - Run2 replay config dropped the write_miniaod statements, they should stay
      (both the default False and the individual overrides for some datasets)

The dropping of dqm_sequences from the prod config should also be done for what we run in prod. I think all datasets affected are not in the current HLT menus, but we should still do it.

lucacopa:
ok, all the Parked PDs are there, you can merge this.

lucacopa:
I updated the configs according to your comments, please have a look. Also, I checked the parked PDs, they are all included.

hufnagel:
Looks good ! One minor change, change the Express GT to GR_E_V49 (once my replay testing it finishes). Then I would propose we do a couple quick replays of a short Run1 run and a short Run2 run to confirm the configs are functional and then we merge it.

hufnagel:
Express with GR_E_V49 has been validated, works fine. PCL is functional now even for older runs.

lucacopa:
ok, I updated the prod config with the latest release too. I will run a short test to confirm that the replay configs are ok too.

PR #4151, opened 2015-06-23
Pass memory estimates through the splitting algorithms

lucacopa:


hufnagel:
Won't be able to test this until next week, but go ahead and test it together with the other patches. Didn't realize this has to be forwarded from the spec through the job splitter into every single job, but given how the JobSubmitter works (only knowing job records), it makes sense.

Question, so merge/logcollect/harvest/cleanup/etc jobs, that don't do this, all get 1250 as default ?

lucacopa:
Yeah, currently the only splitting algorithms that support this are: FileBased, EventBased, LumiBased and EventAwareLumiBased. This patch would add Repack and Express to this list. All the other algorithms in WMCore or the T0 are not passing the memory requirement from the request, so they are all using the 1250 default.

lucacopa:
@hufnagel Patch is working for both Express and Repack jobs

PR #4154, opened 2015-06-30
make sure repack and express workflows wait for streamer notifications

hufnagel:
If StorageManager notifications are enabled, require all streamers to be 'deleted' (SM has been notified) before the repack and express workflows can be marked as injected. For repack workflows we still require also all PromptReco for ddatasets in the run/stream to be released for PromptReco.


hufnagel:
Tested in replays in both modes, seems to be doing what its supposed to (timing of marking workflows as finished looks correct).

PR #4156, opened 2015-07-01
Update tier0 Prod configuration for 1.9.97 deployment

lucacopa:


lucacopa:
@hufnagel please have a look, it always to have more eyes over such changes

hufnagel:
Hard to look at the changes like this, what's the diff to the current Run2 replay configuration. Should be minimal and for the usual suspects. As long as that is the case, this should be fine.

lucacopa:
Difference is basically the production urls and settings (AcqEra, backfill None...) I tested it and it is working fine. Please merge this.

PR #4162, opened 2015-07-19
Add the HighMultiplicity85 dataset

lucacopa:


lucacopa:
Brandon pointed that there were 2 entries for the HighMultiplicity datasets in the config. 
This is a mistake, I am updating one of them to HighMultiplicity85

hufnagel:
Are we sure that the HighMultiplicity dataset was a mistake ? When was it added ?

lucacopa:
The duplicate was added with the 747 release update:
https://github.com/dmwm/T0/commit/1774f35b1f6ebcd61b4bef9fb9f462e9c8d65c4e#diff-49ee9258c3a81a3f0c3f7d62494aeb5f
This is te request we got:
https://hypernews.cern.ch/HyperNews/CMS/get/tier0-Ops/1055.html

hufnagel:
Well, I see a request to add HighMultiplicity and I see another later request to add HighMultiplicity85. What I do not see is that this second dataset superseeds the first. For all I know they could both exist.

hufnagel:
In fact, there is data for both datasets, so we need to have both of them in the config !

hufnagel:
Ah, sorry, looking at the code I see that HighMultiplicity  is actually there in another section, never mind.

PR #4165, opened 2015-07-22
Remove bad exit codes from the failureExitCodes config

lucacopa:


lucacopa:
@hufnagel I deployed an instance with this change, configuration is set correctly.

PR #4166, opened 2015-07-30
Update with the HLTMonitor stream

johnhcasallasl:


hufnagel:
The replay config doesn't match the prod config. The DQMOffline and ppRun2B0T was a one-time thing for the replays, the github replay config should match what we'll deploy in prod eventually.

I in any case will only merge this when we are done with the tests (also because we want to add the dqm_sequences parameter). So I am leaving this open for now and it should be adjusted later to the actual config we'll be using.

hufnagel:
Close this because the change was merged as part of the 25ns config update.

PR #4168, opened 2015-08-07
Remove WorkQueueManager from the Tier0 config

lucacopa:


lucacopa:
@hufnagel This is working correctly, also it is deployed in prod

PR #4172, opened 2015-08-12
Add Skims to the Tier0 configuration

lucacopa:
- Add skims to the Tier0 configuration
- remove comments from dqm sequences
- enable multicore for PromptReco and Express (multicore 4)
- New alca matrix updates
- Default write_dqm = False
- Enabling dqm for PDs with alca producers
- @common default when write_dqm = True
- Ignore DQMOffline stream

### DIFF
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -78,7 +78,7 @@ setPromptCalibrationConfig(tier0Config,
 
 
 # Defaults for CMSSW version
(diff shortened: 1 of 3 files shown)


lucacopa:
Currently testing this

lucacopa:
@hufnagel this can be merged, it is deployed in prod

slava77:
@deguio do we need to specify miniaod dqm as well?

hufnagel:
@slava77 , if there is an issue with miniaod and dqm, can you open a new thread in HN instead of commenting on already merged/closed issues :-) ?

PR #4187, opened 2015-08-26
Setting ppScenario to ppRun2at50ns

johnhcasallasl:


hufnagel:
You forgot to change the Express configuration to use ppScenarioExpress.

PR #4204, opened 2015-09-18
update to 7_4_12_patch1. also updated AcqEra to 2015D

blallen:


blallen:
There are some minor differences between these config files and those used in the production node, mainly the lack of the ParkingMonitor PDs. I manually resolved the differences on the tier0 headnode.

PR #4222, opened 2015-11-16
Adding PDs for HI Run 2015

johnhcasallasl:


hufnagel:
These don't use the ppScenario, but the HeavyIonsRun2 scenario. As there is only a single scenario for these, I would just make a list for all these datasets and their _0T copies and then have a loop adding them for HeavyIonsRuns2 PromptReco.

PR #4237, opened 2015-11-25
create HI replay config

hufnagel:


hufnagel:
@ElianaBohorquez , please take a look whenever convenient.

PR #4255, opened 2016-02-11
AqcEra set to Commissioning 2016

johnhcasallasl:
AqcEra set to Commissioning 2016. Using CMSSW_7_5_8_patch1. HI configs removed from here.


johnhcasallasl:
Just updated the PR to use CMSSW_7_5_8_patch3 instead of CMSSW_7_5_8_patch1

PR #4256, opened 2016-02-23
Changes for MWGR2 2016

johnhcasallasl:
+ New CMSSW version
+ New Architecture
+ Override of architecture for CMSSW_7_4_15
+ New scenarios (config and DB schema)
+ New GTs
+ @common dqm sequence 
+ Setting max latency in the default Repack config (only replays)

### DIFF
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -61,7 +61,7 @@ setDQMDataTier(tier0Config, "DQMIO")
 # First timeout is used directly for reco release
 # Second timeout is used for the data service PromptReco start check
 # (to basically say we started PromptReco even though we haven't)
-defaultRecoTimeout =  48 * 3600 * 100
+defaultRecoTimeout =  48 * 3600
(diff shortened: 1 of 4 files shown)


johnhcasallasl:
- Override of architecture for CMSSW_7_5_8 added too

johnhcasallasl:
Based on what Eliana proposed & David commented on HN:
- Architecture remains in 491
- CMSSW version in 8_0_0_patch1
- Override of architecture for CMSSW_8_0_0_patch1 to use 493. No more overrides needed. 
- Last year scenarios

ebohorqu:
We are missing @common sequence in these datasets: https://github.com/johnhcasallasl/T0/blob/MWGR2_2016/etc/ProdOfflineConfiguration.py#L286

ebohorqu:
Maybe we also should leave "defaultRecoTimeout =  48 * 3600" instead "defaultRecoTimeout =  48 * 3600 * 100"

johnhcasallasl:
Done. Thanks for checking!

PR #4357, opened 2017-06-09
Prod replay config merge

blallen:
Also brings configs up to date with 921, latest alca matrix, and latest global tags.


etc/ProdOfflineConfiguration.py
```diff
@@ -32,6 +32,9 @@
 # Set the verstion configuration (not used at the moment)
 setConfigVersion(tier0Config, "replace with real version")
 
+# replay these run
+#setInjectRuns(tier0Config, [ 289461 ])
```
hufnagel: You really want a replay run in the production config ?

etc/ProdOfflineConfiguration.py
```diff
@@ -222,10 +205,7 @@
            sizePerEvent = 4000,
            scenario = ppScenario)
 
-datasets = [ "Commissioning1_0T", "Commissioning2_0T", "Commissioning3_0T", "Commissioning4_0T" ]
-
-for dataset in datasets:
-    addDataset(tier0Config, dataset,
+    addDataset(tier0Config, dataset+"_0T",
```
hufnagel: If you restructure the config like this, can you please do it for all datasets ?

etc/ProdOfflineConfiguration.py
blallen: I think this was mostly intended as a reminder, but it might not be necessary now we have synced the replay config. @johnhcasallasl , what do you think?

etc/ProdOfflineConfiguration.py
blallen: Yeah, I can do it. It was just taking a while, so I only did it to the datasets I was already restructuring for other reasons.

PR #4366, opened 2017-07-14
Removed the lines of the setExpressSubscribeNode import and its usage.

vytjan:


hufnagel:
I am not going to merge this until we actually switch the production Tier0 to use this. If you want to change the PR to only make this change for the replay configurations that would be acceptable. But the production config in GH should match what we are using in the production Tier0 instance.

PR #4370, opened 2017-08-08
Prod config updated for the 2.0.8 release.

vytjan:
Current production configuration on the vocms014


hufnagel:
The Express config for Calibration and ExpressAlignment make no sense. Not going to merge this. And if we actually run this it should be changed ASAP. As-is none of the data that is produced by these workflows goes anywhere. You need to define what to do with it (ie. subscriptions) !

hufnagel:
Actually, I'll merge it, but this should be changes ASAP. Will write to mailing list.

PR #4372, opened 2017-08-22
Add ErrorHandler failure exit codes to T0 config

amaltaro:
This attribute was removed from the standard WMAgent configuration in (as of 1.1.6.pre6):
https://github.com/dmwm/WMCore/pull/8050

@hufnagel I haven't added the new exit codes CompOps was asking to be added.

### DIFF
--- a/etc/Tier0Config.py
+++ b/etc/Tier0Config.py
@@ -36,4 +36,6 @@ config.DBS3Upload.primaryDatasetType = "data"
 
 config.DBSInterface.primaryDatasetType = "data"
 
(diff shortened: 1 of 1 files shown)


hufnagel:
Why ? Jobs with these exit codes will go straight to failed, without being paused. Is that what we want ?

amaltaro:
Yes, and that's what you always had from the standard WMAgent config. Unless you were removing it at some point in the T0 agent deployment.
The point is, it has just been removed from the agent config, so if you were relying on that, please keep it in the T0 config.

hufnagel:
tier0-mod-config sets it to []

amaltaro:
I see, thanks!

PR #4382, opened 2017-09-11
Updated alca matrix. Created new configuration file.

cmst0:
- Removed LumiPixels from  alca_producers for AlCaLumiPixels. Updated defaultRecoSplitting. 
- New ProdOfflineConfiguration for future deployments. This will include possible changes for production.


hufnagel:
Who is cmst0 ? We shouldn't use generic account for pull requests IMO as this makes the github history useless. I want a name associated with a change.

As a functional comment, please collapse the changes into one commit. Otherwise we pollute the repository history a lot.

PR #4383, opened 2017-09-13
update PromptReco and Express memory

hufnagel:


drkovalskyi:
If I read it correctly we have implement (N-1) instead of N in the formula suggested by Slava, i.e.3500 + 1050*N,

hufnagel:
No, it's 3500 + N*1050. I add one extra first, then the N-1.

PR #4387, opened 2017-09-23
Revert to previous values to limit Express input rate.

blallen:
Changes I mentioned on HN: https://hypernews.cern.ch/HyperNews/CMS/get/tier0-Ops/1720/1/1/1.html

maxInputEvents is 400 because that's what we had before using the workaround. Not sure why it's different from the maxInputRate. Do these two parameters need to be the same?


hufnagel:
maxInputEvent determines how many events a single express job is allowed to process (express jobs in principle are multi-lumi). With it you can only force it do go down to a single lumi per job, you cannot fail lumis.

PR #4396, opened 2017-11-14
Config clean up and ppRef 2017

blallen:
Reorganized current configs so PDs are in logical order. Most actual deletions are removing the 0T datasets. For details of new organization and what can probably be removed in a future PR, see HN thread: https://hypernews.cern.ch/HyperNews/CMS/get/tier0-Ops/1750/1/1/1/1.html

Also contains all the changes made for the 2017 ppRef run. Currently matches the production config in use on vocms0313.


hufnagel:
20 commits ???

drkovalskyi:
What's the problem with that? Is there a limit on commits?

hufnagel:
It pollutes the commit history, you have a hard time figuring out what's going on. And yes, multiple commits are allowed/encouraged for functionally separate changes.

But things like  "merge updates from master" and "merge pull request from..." are just complete junk.

@blallen , can you clean this up / rebase ?

hufnagel:
For example, in WMCore the policy is that one PR has normally one commit and another optional one for pylint style changes. Multiple functional commits are possible if the PR bundles some independent functional changes.

drkovalskyi:
I don't see how collapsing all commits in one helps to figure out what's going on. While I can agree that some commits can be merged together (like those that represents trivial formatting changes), but I still don't see what exactly it pollutes.
If WMCore has some rules against frequent commits, let's fork or use a different repo for configuration and you will pull whatever you need for WMCore, but we will continue using the fork for ops.

hufnagel:
There is no technical limit here, just a sensible policy. Which it falls to me to enforce as the owner of the T0 repository.

If you want to keep the functional commits, fine. I don't like it personally (too fine-grained for no reason IMO), but at least remove/rebase the technical ones that have no purpose.

hufnagel:
Sure, but don't bundle 20 at a time and make a single PR two weeks later. Anyways, I have given up on the functional commits, leave them as-is. Just rebase away the crap and I'll merge it.

hufnagel:
In addition, most of the commits are from @blallen , so that really doesn't improve the "blame" game...

hufnagel:
For the future, if a user makes a functional change, that user should make a PR. If there are multiple independent functional changes (where independent means they could have been merged individually and things would still work with just a subset), from the SAME user, by all means, include multiple commits in the same PR.

Don't group commits from multiple users over an extended period of time and then make a combined PR.

And if your argument for grouping is: well, these are related and we were restructuring everything at once, then I would say these aren't really independent commits anymore and you should have made a single restructuring commit.

drkovalskyi:
I don't think we care what user made a change. What we need to know is the reason for the change, i.e. we need meaningful comments to be able to track changes and understand why things were changed if there is any doubt if it's still needed or not.

Let's do the following. Brandon will cleanup the commits and reduce them for this case, but let's discuss it in details in the coordination meeting to decide how we want to do it in future taking into account the need to integrate it with tools such as jenkins.

blallen:
I made a new branch and PR with all the changes in a single commit: https://github.com/dmwm/T0/pull/4397

PR #4477, opened 2019-03-05
Add Apache 2 license and notice

andresfelquintero:
Added Apache 2 license and notice as WMCore repository did.


vytjan:
@andresfelquintero AFAIK, the LICENSE and the NOTICE files should not have any extension (as they are .txt now).

PR #4478, opened 2019-03-05
Add Apache 2 License and Notice

andresfelquintero:
Added Apache 2 license and notice as WMCore

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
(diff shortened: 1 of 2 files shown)


vytjan:
@amaltaro , @hufnagel could one of you confirm if only the LICENSE and NOTICE files are needed?
As I recall from the last DMWM meeting, there is something else needed with the licenses, right?

amaltaro:
This is the most important as far as I could see. In the future, we can add a short header to every single source code, but we're not going to do so in WMCore, at least not for the moment.

vytjan:
Thanks for the feedback Alan. I believe T0 will just follow what you guys are doing for WMCore (talking about that short header in every source code file).

PR #4484, opened 2019-03-15
Unit_test_ExpressMerge_t

andresfelquintero:
I added some modules that were missing from the configuration.


vytjan:
Andres, it's 3 different commits here - could you please squash them into a single commit for this PR?

PR #4499, opened 2019-05-16
Updated Tier0Feeder_t

andresfelquintero:
added an additional line that was missing from the configuration.


vytjan:
Thanks for the fix Andres. 
Yet another example why we need to have properly working both unit tests and code quality checks.

PR #4500, opened 2019-06-03
Updated ConditionUploadAPI.py

andresfelquintero:
Changed the input as suggested by Alan Malta.


vytjan:
Andres, originally at #249 the "lines" variable is a list of all the lines of a file ( fin.readlines() ) and it should be kept so as we don't want to change the logic of the code. Could you please change that line back to what it was?
The rest seems to be fine.

PR #4501, opened 2019-06-06
Updated conditionuploadAPI.py

andresfelquintero:
I did the changes suggested by Vytas.


vytjan:
Tested it in a replay, it works fine.

PR #4502, opened 2019-06-07
Adding a configurable number of retries default configuration. More i…

vytjan:
…nformation on T0 twiki cookbook.


vytjan:
Tested in multiple replays, works fine.

PR #4507, opened 2019-08-01
Do not allow T0 to use WorkQueue in any case.

vytjan:
A mandatory patch for T0 2.1.5 release.


vytjan:
Let's keep the up-to-date mandatory T0 patches here, not randomly distributed on different Tier0 VMs.

PR #4521, opened 2020-01-17
Mwgr#0 config

germanfgv:
* updated_offline_config
I updated the CMSSW version as requested.

* Updated ReplayOfflineConfiguration for MWGR#0
Updated CMSSW version and CMSSW override.

* Updated ReplayOfflineConfiguration for MWGR#0
Updated CMSSW version and CMSSW override.


germanfgv:
Ok. I'm going to fix the min run number and the make another request

PR #4532, opened 2020-08-11
Test WMCore PR 9859

drkovalskyi:
Infrastructure testing

### DIFF
--- a/etc/ContainterConfig.sh
+++ b/etc/ContainterConfig.sh
@@ -3,4 +3,4 @@ T0_VERSION=2.2.0
 
 #WMC_PR specifies a list of WMCore Pull Request to be tested
 #WMC_PR=""
-WMC_PR=""
+WMC_PR="9859"


germanfgv:
test this please

drkovalskyi:
test this please

drkovalskyi:
@germanfgv what's the status of this PR test?

germanfgv:
It had 2 paused jobs due to memory usage. All the other jobs finished successfully.

drkovalskyi:
Well it means it's not the best run to test on. Why don't we use a collision run from 2018?

drkovalskyi:
Also, let's introduce a bunch of options of what kind of replays we can trigger, i.e. pp 2018, HI 2018, cosmics 2020, mwgr2 2020 etc - something along that line.

germanfgv:
I agree it would be best if the user can request a different run by specifying which type of replay to test. Right now that can only be done by modifying ReplayOfflineConfiguration.py on the PR commit. I'll add the option to ContainerConfig.py. I guess we should also add GTs and CMSSW version as parameters.

drkovalskyi:
While specifying a run can be a good feature, what I'm after is a more generic feature. As a developer I want to test my code for a specific configuration, not a specific run in most cases, i.e. we should have standard tests where we set default run values and then just say "test pp2018" or "test hi2018" or "test latest".
Btw, look at https://github.com/dmwm/WMCore/pull/9859 and see how many different tests they are running - we need the same.

gkfthddk:
commenting test

gkfthddk:
commenting test

gkfthddk:
commenting test

gkfthddk:
commenting test

gkfthddk:
commenting test

PR #4557, opened 2021-02-18
Replay test with 11_2_2

silviodonato:
This PR is to try the automatic replay.
Here I'm using 11_2_2. 
I will use 11_2_2_patch1 for the actual replay in few hours.

### DIFF
--- a/etc/ReplayOfflineConfiguration.py
+++ b/etc/ReplayOfflineConfiguration.py
@@ -90,7 +90,7 @@ setPromptCalibrationConfig(tier0Config,
 
 # Defaults for CMSSW version
 defaultCMSSWVersion = {
-       'default': "CMSSW_11_2_2"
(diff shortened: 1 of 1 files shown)


silviodonato:
ah, probably I have to use https://github.com/CMSTier0

drkovalskyi:
You did it all correctly

silviodonato:
> Can one of the admins verify this patch?

@drkovalskyi I was confused by this message.
Perhaps it needs some time to see that I'm member of https://github.com/CMSTier0

germanfgv:
Hi @silviodonato. Everthing seems fine. I'm also a bit confused by that comment, but maybe you are right and it just needs time. Please, go ahead and request the replay when the new release is ready. I'll take a look at our Jenkins config to make sure everything is fine.

PR #4571, opened 2021-05-04
Replay test with CMSSW_11_3_0_pre6

silviodonato:
First test of CMSSW_11_3_X for the MWGR#4 (2-4 June).
There will be a CMSSW_11_3_0 ~next week, but it would be good to test CMSSW_11_3_X in advance, but I'm not sure if we can make a test with a prerelease.
I've also updated the global tag to https://github.com/cms-sw/cmssw/blob/master/Configuration/AlCa/python/autoCond.py

PR #4572, opened 2021-05-07
Replay test with CMSSW_11_3_0

silviodonato:
Test of CMSSW_11_3_0 for the MWGR#4 (2-4 June).
I've also updated the global tag to https://github.com/cms-sw/cmssw/blob/master/Configuration/AlCa/python/autoCond.py

In the previous test #4571 - based on CMSSW_11_3_0_pre6 - there was an issue with the global tag.
The problem should be solved in 11_3_0 (see https://github.com/dmwm/T0/pull/4571#issuecomment-832228457)

### DIFF
--- a/etc/ReplayOfflineConfiguration.py
+++ b/etc/ReplayOfflineConfiguration.py
@@ -136,7 +136,7 @@ repackVersionOverride = {
     "CMSSW_11_1_5" : defaultCMSSWVersion['default'],
     "CMSSW_11_2_1" : defaultCMSSWVersion['default'],
     "CMSSW_11_2_2" : defaultCMSSWVersion['default'],
-    "CMSSW_11_2_3" : defaultCMSSWVersion['default']
+    "CMSSW_11_2_3" : defaultCMSSWVersion['default'],
     "CMSSW_11_2_4" : defaultCMSSWVersion['default']
     }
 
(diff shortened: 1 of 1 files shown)


silviodonato:
run replay please

germanfgv:
There was a paused job due a SYSTEM_PERIODIC_REMOVE in condor. While I'm looking into exactly why this happened, I resumed the jobs, since this kind of issue is usually temporary in nature. The replay is still running without issues.

We need to fix the reports of our monitoring script to avoid multiple notifications because the same paused job.

germanfgv:
@silviodonato This time there is a real error with the replay. The error message we are getting is the following:
```
INFO:root:addOutputFile method called with outputModule: write_DQMIO, aFile: None
INFO:root:addOutputFile method fileRef: , whole tree: {}
CRITICAL:root:Error running cmsRun
{'arguments': ['/bin/bash', '/srv/job/WMTaskSpace/cmsRun1/cmsRun1-main.sh', '', 'slc7_amd64_gcc900', 'scramv1', 'CMSSW', 'CMSSW_11_3_0', 'FrameworkJobReport.xml', 'cmsRun', 'PSet.py', '', '', '']}
CMSSW Return code: 8002

CRITICAL:root:Error message: An exception of category 'StdException' occurred while
   [0] Processing  Event run: 341169 lumi: 517 event: 9310078 stream: 7
   [1] Running path 'dqmoffline_step'
   [2] Calling method for module CSCTriggerPrimitivesProducer/'valCscStage2Digis'
Exception Message:
A std::exception was thrown.
vector::_M_range_check: __n (which is 96) >= this->size() (which is 96)
```
I'll also report this error in HN.

silviodonato:
moved to https://github.com/dmwm/T0/pull/4573

PR #4575, opened 2021-06-01
Change T0 code from py2 to py3

jhonatanamado:
T0 code using py3


jhonatanamado:
I will close this PR. Please use #4603 to follow some other changes introduced in the T0 code.

PR #4584, opened 2021-07-07
Paused failing jobs

jhonatanamado:
While testing new WMCore releases with replays, we need to ensure that all types of CMS jobs finished successful instead of go failing after x attempts (for example for LogCollect and CleanUp jobs). The Tier0Config.py file shows the [following](https://github.com/dmwm/T0/blob/fa9a200af6d492a80beefe158afcad9015e4d630/etc/Tier0Config.py#L23-L29) logic so we only need to overwrite this configuration only for replays.


germanfgv:
@jhonatanamado You made this changes on top of the py2 topy3 migration changes. Could you please separate the two?

drkovalskyi:
22 files changed?! Let's try to make finer changes to be able to make sense of what and why has been changed.

PR #4590, opened 2021-07-20
Tunning LogCollect and CleanUp Jobs

jhonatanamado:
With the current configuration, LogCollect and CleanUp jobs go to fail state after the second attempt. Changing RetryManager and ErroHandler to test new versions of T0 for only replays.


germanfgv:
I see no problems with the changes. @jhonatanamado I'm guessing you already tested the functionality of this changes, right?

jhonatanamado:
Yes, the changes were tested trying to build new releases for T0, and to get the logs, tarballs and sandboxes for LogCollect jobs that were failing automatically this lines must to be added in order to debug failures and have a good release of our agent

PR #4591, opened 2021-07-20
Configurable StreamMemory

jhonatanamado:
Set the stream memory configurable. Default value will be 1000 unless is assigned using in the OfflineConfig file the following function:
`setStreamMem(tier0Config,"NanoDST",value=2000)`

### DIFF
--- a/etc/ReplayOfflineConfiguration.py
+++ b/etc/ReplayOfflineConfiguration.py
@@ -23,7 +23,7 @@ from T0.RunConfig.Tier0Config import addExpressConfig
 from T0.RunConfig.Tier0Config import setInjectRuns
 from T0.RunConfig.Tier0Config import setStreamerPNN
 from T0.RunConfig.Tier0Config import setEnableUniqueWorkflowName
-
+from T0.RunConfig.Tier0Config import setStreamMem
 # Create the Tier0 configuration object
 tier0Config = createTier0Config()
 
@@ -1516,6 +1516,11 @@ ignoreStream(tier0Config, "streamHLTRates")
 ignoreStream(tier0Config, "streamL1Rates")
 ignoreStream(tier0Config, "streamDQMRates")
 
+###################################
+### configure memory for stream ###
+###################################
+setStreamMem(tier0Config,"NanoDST",value=2000)
+setStreamMem(tier0Config,"ScoutingPF",value=2000)
 ###################################
 ### currently inactive settings ###
 ###################################
--- a/src/python/T0/RunConfig/RunConfigAPI.py
+++ b/src/python/T0/RunConfig/RunConfigAPI.py
@@ -22,6 +22,7 @@ from WMCore.ReqMgr.DataStructs.RequestStatus import REQUEST_START_STATE
 from T0.RunConfig.Tier0Config import retrieveDatasetConfig
 from T0.RunConfig.Tier0Config import addRepackConfig
 from T0.RunConfig.Tier0Config import deleteStreamConfig
+from T0.RunConfig.Tier0Config import setStreamMem
 
 from WMCore.WMSpec.StdSpecs.Repack import RepackWorkloadFactory
 from WMCore.WMSpec.StdSpecs.Express import ExpressWorkloadFactory
@@ -382,7 +383,7 @@ def configureRunStream(tier0Config, run, stream, specDirectory, dqmUploadProxy):
 
             streamConfig.Express.ScramArch = tier0Config.Global.ScramArches.get(streamConfig.Express.CMSSWVersion,
                                                                                 tier0Config.Global.DefaultScramArch)
-            
+
             streamConfig.Express.RecoScramArch = None
             if streamConfig.Express.RecoCMSSWVersion != None:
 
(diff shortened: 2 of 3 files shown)


src/python/T0/RunConfig/Tier0Config.py
```diff
@@ -573,6 +573,18 @@ def setProcessingSite(config, site):
...
     return
 
+def setStreamMem(config, streamname, **options):
+    """
+    _SetStreamMem_
+    Define memory usage for any type of streamer configurable.
+    The following keys may be passed:
+        value - size of memory to be assigned 
+    """
+    if 'value' in options:
```
germanfgv, line 583: I don't understand why the method receives an unespecified number of optional values. Do we need something other than 'value' ? is there a use case in which we wouldn't provide 'value'?. I think we could just use:
```
def setStreamMem(config, streamname, value):
```
or expand the scope of the funciont and make it a tool to define the full streamer configuration:
```
def addStreamerConfig(config, streamname, value):
```

src/python/T0/RunConfig/Tier0Config.py
```diff
@@ -573,6 +573,18 @@ def setProcessingSite(config, site):
...
+        config.Global.StreamMem = options['value']
```
germanfgv, line 584: You are changing `StreamMem` for the global configuration. This means you are not really allowing us to configure different memory values for different streamers, just changing the default value.

src/python/T0/RunConfig/RunConfigAPI.py
```diff
@@ -553,18 +554,18 @@ def configureRunStream(tier0Config, run, stream, specDirectory, dqmUploadProxy):
...
+                    tier0Config.Global.AcquisitionEra, streamConfig.Repack.ProcessingVersion,
                     time.strftime('%y%m%d_%H%M', time.localtime(time.time())))
             else:
                 workflowName = "Repack_Run%d_Stream%s" % (run, stream)
 
             specArguments = {}
 
-            if stream == 'ScoutingPF':
-                specArguments['Memory'] = 2000
+            if stream == setStreamMem(tier0Config,stream):
```
germanfgv, line 565: with the current deffinition of `setStreamMem`, this will always evaluate to `True`. I don't think that is what you are trying to achieve.

src/python/T0/RunConfig/RunConfigAPI.py
```diff
@@ -553,18 +554,18 @@ def configureRunStream(tier0Config, run, stream, specDirectory, dqmUploadProxy):
...
+                specArguments['Memory']=tier0Config.Global.StreamMem
```
germanfgv, line 566: All streamers will end up with `Global.StreamMem`

germanfgv (commented):
As Dima mentioned on the Jira ticket, I think you should look `addRepackConfig` to see how to handle independent configurations for different streamers. the current code does not achieve the goal of allowing different memory requirements for each stream.

src/python/T0/RunConfig/Tier0Config.py
```diff
@@ -1020,3 +1020,17 @@ def addTier1Skim(config, skimName, dataTier, primaryDataset, cmsswVersion,
 
     datasetConfig.Tier1Skims.append(skimConfig)
     return
+
+def setStreamMem(config, *streamid, **streamidvalue):
```
germanfgv, line 1024: I'm not sure why need *streamid to acomplish this. What is the use of having dictionary keys named "arg1", "arg2", etc.? there are never going to bwe recognized in RunConfigAPI

src/python/T0/RunConfig/Tier0Config.py
```diff
@@ -1020,3 +1020,17 @@ def addTier1Skim(config, skimName, dataTier, primaryDataset, cmsswVersion,
...
+    _setStreamMem_
+    For each streamid and streamidvalue in KB, create a dict to update the max
+    memory within the specArgument['Memory']
+    """
+
+    streamidmem = {}
+    for i, a in enumerate(streamid):
+        streamidmem['arg%s' % (i+1)] = a
+    streamidmem.update(streamidvalue)
+    config.Global.StreamMem = streamidmem
```
germanfgv, line 1035: I was thinking that you were going to use our already existing streamer configs. I think this approach could work, but I'm not sure if it is the best to have an extraset of streamer configurations.

src/python/T0/RunConfig/RunConfigAPI.py
```diff
@@ -561,9 +562,10 @@ def configureRunStream(tier0Config, run, stream, specDirectory, dqmUploadProxy):
...
 
             specArguments = {}
+            streamidmem = dict(tier0Config.Global.StreamMem)
 
-            if stream == setStreamMem(tier0Config,stream):
-                specArguments['Memory']=tier0Config.Global.StreamMem
+            if stream in streamidmem:
+                specArguments['Memory'] = streamidmem.get(stream)
             else:
                 specArguments['Memory']=1000
```
germanfgv, line 570: New default should be 2000

etc/ReplayOfflineConfiguration.py
```diff
@@ -1519,8 +1519,7 @@
 ###################################
 ### configure memory for stream ###
 ###################################
-setStreamMem(tier0Config,"NanoDST",value=2000)
-setStreamMem(tier0Config,"ScoutingPF",value=2000)
+setStreamMem(tier0Config, NanoDST = 2000, ScoutingPF = 2000)
```
germanfgv, line 1522: As I noted in Tier0config.py, you are not using *streamid. Maybe you could explain the use case that this list covers.

germanfgv (commented):
This solution should work, althouhg with some changes. Nevertheless, I'm not sure it is better that simply adding another attribute to our streamer configs.

PR #4595, opened 2021-08-04
ConfigurableStreamMemory

jhonatanamado:
Set the memory of streams configurable via addRepackConfig

### DIFF
--- a/src/python/T0/RunConfig/Tier0Config.py
+++ b/src/python/T0/RunConfig/Tier0Config.py
@@ -65,7 +65,7 @@ Tier0Configuration - Global configuration object
 |       |
 |       | |--> Default - Configuration section for repacking defaults
 |       |                (configured just like a repack stream)
-|       |      
+|       |
 |       |--> STREAMNAME - Configuration section for a stream.
 |             |
 |             |--> ProcessingStyle - The processing style for the stream
@@ -127,7 +127,7 @@ Tier0Configuration - Global configuration object
 |             |     |--> ProcessingVersion - processing version
 |             |     |
 |             |     |--> MaxInputRate - max input rate that is accepted for processing
-|             |     |                       (in events per lumi section) 
+|             |     |                       (in events per lumi section)
 |             |     |
 |             |     |--> MaxInputEvents - max input events for express processing job
 |             |     |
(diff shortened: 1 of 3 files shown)


drkovalskyi:
@jhonatanamado I don't think we need setConfStreamMemory. Function `addRepackConfig` requires to specify the stream where configuration is applied. In other words we set the default with `addRepackConfig(tier0Config, "Default", ...)` and modify configuration for a specific stream with extra calls of `addRepackConfig` specifying the stream name explicitly. Check how we use `addDataset`.

Please clean up the pull request no to have changes where they are not needed, i.e. white spaces, new lines etc.

src/python/T0/RunConfig/Tier0Config.py
```diff
@@ -789,6 +789,15 @@ def addRepackConfig(config, streamName, **options):
     else:
         streamConfig.Repack.BlockCloseDelay = options.get("blockCloseDelay", 24 * 3600)
 
+    if hasattr(streamConfig.Repack, "SetStreamMemory"):
+        streamConfig.Repack.SetStreamMemory = options.get("setStreamMemory", streamConfig.Repack.SetStreamMemory)
+    else:
+        streamConfig.Repack.SetStreamMemory = options.get("setStreamMemory", 2000)
+
+    if hasattr(streamConfig.Repack, "SetConfStreamMemory"):
+        streamConfig.Repack.SetConfStreamMemory = options.get("setConfStreamMemory", streamConfig.Repack.SetConfStreamMemory)
```
germanfgv, line 798: As Dima already commented, this attribute is not necessary. We don't need a list of streams to modify, as what we will do is add a new configuration for each stream that needs a configuration different from the default.

src/python/T0/RunConfig/Tier0Config.py
```diff
@@ -789,6 +789,15 @@ def addRepackConfig(config, streamName, **options):
     else:
         streamConfig.Repack.BlockCloseDelay = options.get("blockCloseDelay", 24 * 3600)
 
+    if hasattr(streamConfig.Repack, "SetStreamMemory"):
```
germanfgv, line 792: I would change the name of this attribute, as this name suggest an action (reads like a method's name). I would use something like `"MaxMemory"` instead.

germanfgv (commented):
This is a lot closer to what we need, but it does need a few changes.

PR #4566, opened 2021-03-09
Using WMWorkloadHelper.setTaskEnvironmentVariables to override site config

germanfgv:
Uses changes introduced in https://github.com/dmwm/WMCore/pull/10344 to override the site configuration file used by T0 jobs

PR #4596, opened 2021-08-04
Add sites configuration to easily switch storage sites

germanfgv:
Before #4566, choosing a particular processing site implied that we were using the same site as storage. Now we have the ability to process data in T2_CH_CERN while storing output in different disk sites. This feature is currently being use in production to write output data to T0_CH_CERN_Disk and to test EOS configurations with EOS_PILOT.

Given this development, our previous way of configuring sites needed to be reworked. This PR does that by adding the following fields to the Tier0 configuration:

- Added StorageSite as a Global configuration parameter aside from ProcessingSite: This makes it more clear to pick different storage sites.  

- Added Site configurations: Now we can specify relevant information when using a new site

Now  TFC and site local config files are overriten only when there is a difference between ProcessingSite and StorageSite

### DIFF
--- a/etc/ReplayOfflineConfiguration.py
+++ b/etc/ReplayOfflineConfiguration.py
@@ -20,8 +20,10 @@ from T0.RunConfig.Tier0Config import setConfigVersion
 from T0.RunConfig.Tier0Config import ignoreStream
 from T0.RunConfig.Tier0Config import addRepackConfig
 from T0.RunConfig.Tier0Config import addExpressConfig
+from T0.RunConfig.Tier0Config import addSiteConfig
 from T0.RunConfig.Tier0Config import setInjectRuns
 from T0.RunConfig.Tier0Config import setStreamerPNN
+from T0.RunConfig.Tier0Config import setStorageSite
 from T0.RunConfig.Tier0Config import setEnableUniqueWorkflowName
 
 # Create the Tier0 configuration object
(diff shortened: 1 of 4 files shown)


etc/ReplayOfflineConfiguration.py
```diff
@@ -35,8 +37,19 @@
 
 # Settings up sites
 processingSite = "T0_CH_CERN"
```
jhonatanamado, line 39: Hi @germanfgv , I believe this should be T2_CH_CERN

etc/ReplayOfflineConfiguration.py
```diff
@@ -35,8 +37,19 @@
...
+storageSite = "T0_CH_CERN_Disk"
 streamerPNN = "T0_CH_CERN_Disk"
 
+addSiteConfig(tier0Config, "T0_CH_CERN_Disk",
+                siteLocalConfig="/cvmfs/cms.cern.ch/SITECONF/T0_CH_CERN/JobConfig/site-local-config.xml",
```
jhonatanamado, line 44: Due that this siteLocalConfig and overrideCatalog have a default value within the function addSiteConfig, Is it mandatory to use it again here in the OfflineConfig file??

jhonatanamado (commented):
I added some minor comments. The only mandatory thing to change is the value for processingSite. Please add the correct value and update the ProducitonOfflineConfig file :)

etc/ReplayOfflineConfiguration.py
germanfgv: Yes, it is not mandatory. But I think is worth adding it explicitly, so it is clear to everyone what values we are using. Like when we add a default Repack config or default dataset config.

etc/ReplayOfflineConfiguration.py
germanfgv: You are right! I fixed it

jhonatanamado:
@drkovalskyi , I tested this PR with a replay, checked the changes and everything looks good. Any comments ?

drkovalskyi:
Looks reasonable to me. Can we fix the checks so that we don't have failed test before we merge?

germanfgv:
run unit test please

germanfgv:
@drkovalskyi there was a few issues with the unit test set up. I fixed them. Now it is working and makes more sense. We can talk about them tomorrow. For now we can merge this PR.

PR #4603, opened 2021-09-20
Migrating T0 code from PY2 to PY3

jhonatanamado:
This is the last version of T0 code PY3 compatible that was used during several test with CMSSW 12 and WMCore agent 1.5.2


germanfgv:
Everything looks fine @jhonatanamado. Please point to the changes in the cmsdist repo that were needed to properly build this release.

germanfgv:
@jhonatanamado I think you can proceed with the merge.  We will need to fix the Syntax check so it works with the new Python3 agent.

PR #4602, opened 2021-09-03
AlCaDB test of PCL workflows

francescobrivio:
The goal of this PR is:
 - to test the PCL workflows already existing in Run2 by replaying 2018 data 
 - to validate 2 new AlCaRecos `SiPixelCalSingleMuonLoose`, `SiPixelCalSingleMuonTight` which should replace `SiPixelCalSingleMuon` and save some space.

Technical details:
 - Run: 317696 (2018 pp - 4.5h long)
 - CMSSW_12_0_0
 - GTs
   - Express: 113X_dataRun3_Express_Candidate_2021_06_21_16_58_57
   - Prompt: 113X_dataRun3_Prompt_Candidate_2021_06_21_16_59_39
 - The [Prompt Matrix](https://twiki.cern.ch/twiki/bin/view/CMS/AlCaRecoMatrix#Prompt_Configuration_AN7) only includes the `SiPixelCalSingleMuon` and the newly implemented `SiPixelCalSingleMuonLoose/MuonTight`

The rest of the configuration is copied from https://github.com/dmwm/T0/pull/4581 and was implemented by @germanfgv

PR #4604, opened 2021-09-21
Update ExpressConfig to current Tracking Only DQM sequence

jfernan2:
Update following 2019 changes on:
https://github.com/cms-sw/cmssw/blob/master/DQMOffline/Configuration/python/autoDQM.py#L21


germanfgv:
@jfernan2  thank you for your contribution. We will test this DQM sequence in #4602

Issue #4605, opened 2021-09-22
Replay test with 12_1_0_pre3


qliphy:
As discussed with @germanfgv @drkovalskyi  and at ORP meeting yesterday, we (@perrotta and me) will launch replay test with 12_1_0_pre3.  We would like to collect information for that.
@francescobrivio  @tvami @boudoul @rappoccio

Runs  (default setting): 
- 341169,341754,338628,338714,342154 

GTs:   (following https://github.com/dmwm/T0/pull/4602)
- Express: 113X_dataRun3_Express_Candidate_2021_06_21_16_58_57
- Prompt: 113X_dataRun3_Prompt_Candidate_2021_06_21_16_59_39

Others: ?

germanfgv:
@francescobrivio  Do you think it would make sense to wait for 120X GTs?. Also, I'll suggest removing 338714, as I don't think we will benefit of processing 2020 data but it would significantly increase the consumption of resourses. and lenght of the replay.

francescobrivio:
> @francescobrivio Do you think it would make sense to wait for 120X GTs?. Also, I'll suggest removing 338714, as I don't think we will benefit of processing 2020 data but it would significantly increase the consumption of resourses. and lenght of the replay.

@germanfgv 
- I agree with the removal of 338714
- I don't think it's beneficial to wait for 120X GTs, we will need ~1 week for making those GTs and they will be in any case tested once the new release for CRAFT/PilotBeam will be created

qliphy:
@francescobrivio @germanfgv Thanks!
I have made a PR:  https://github.com/dmwm/T0/pull/4607
let me know if you have any further comment before we launch the test.

germanfgv:
@francescobrivio I just noticed that we have not been asked to save the run 344518. Is that right or did I missed something?

francescobrivio:
> @francescobrivio I just noticed that we have not been asked to save the run 344518. Is that right or did I missed something?

@germanfgv good point! thanks for checking! 
I'm looking at https://twiki.cern.ch/twiki/bin/view/CMSPublic/CompOpsTier0TeamReplayGoodRuns but I don't see CRUZET 2021 runs, can you list the runs that have been saved from CRUZET? possibily we should use one that include all subdetectors.

germanfgv:
@francescobrivio You are right. We'll update the list. In the mean time these are the available CRUZET runs:
```
               343498, #CRUZET
[...]

PR #4607, opened 2021-09-23
 Replay test with 12_1_0_pre3

qliphy:
See #4605 

Runs: 341169,341754,342154,344518

Release: CMSSW_12_1_0_pre3

GT: 
- Express: 113X_dataRun3_Express_v4
- Prompt: 113X_dataRun3_Prompt_v3

### DIFF
--- a/etc/ReplayOfflineConfiguration.py
+++ b/etc/ReplayOfflineConfiguration.py
@@ -33,7 +33,7 @@ tier0Config = createTier0Config()
 setConfigVersion(tier0Config, "replace with real version")
 
 # Set run number to replay
-setInjectRuns(tier0Config, [341169,341754,338628,338714,342154])
+setInjectRuns(tier0Config, [341169,341754,342154,344518])
 
 # Settings up sites
 processingSite = "T2_CH_CERN"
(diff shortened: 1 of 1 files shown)


qliphy:
@germanfgv   I noticed "Syntax Check" and "Unit Tests" above show failure message but not obvious details.  Can I just ignore them?

germanfgv:
@qliphy we are still updating our test to run with python3. They still don't work perfectly with our new agent. So yes, they can be ignored. I'll manually the replay and let you know about the results.

etc/ReplayOfflineConfiguration.py
```diff
@@ -100,7 +100,7 @@
 
 # Defaults for CMSSW version
 defaultCMSSWVersion = {
-    'default': "CMSSW_11_3_2"
+    'default': "CMSSW_12_1_0_pre3"
```
francescobrivio, line 103: @qliphy shouldn't you also add the cmssw override for repacking (line 145) and express (line 159)?

qliphy: @francescobrivio   Thanks. Now added.

germanfgv:
@qliphy @francescobrivio after this changes were made, we restarted the replay with the proper config. Around 50% of the jobs have finished with no issues so far.

germanfgv:
As reported in [HyperNewas](https://hypernews.cern.ch/HyperNews/CMS/get/tier0-Ops/2266.html?inline=-1), the replay was successfull with the current configuration. I'll merge this as the current default replay config for Tier0.

PR #4632, opened 2021-10-29
Adding deployment ID to DB and Tier0Config

germanfgv:
Add ID for each T0 deployment. 

The ID is the time and date of the deployment wiht the following format: ''%y%m%d%H%M%S"

The ID is stored in a new table in the T0AST database.  The table preventscratend multiple IDs unless the DB is wiped out.

The deployment ID is also added to the Unique workflow name in replays


jhonatanamado:
Hi German. I tested this PR with multiple replays and everything looks good. Feel free to merge it against master branch

PR #4634, opened 2021-11-12
Update production node CMSSW_12_0_3_patch2

jhonatanamado:
Update the production node with a new CMSSW release and a fix in dqm sequences for MinimumBias PDs


germanfgv (approved):
Looks good

PR #4636, opened 2021-11-24
Add siteLocalConfig to LogCollect overrides

germanfgv:
Configure output location of LogCollect jobs using siteLocalConfig defined in StorageSite
Needs to be deployed with the [WMCore PR#10905](https://github.com/dmwm/WMCore/pull/10905)


germanfgv:
We are already using this in Production and Replays for a while. It works as intended. I'll merge the PR.

PR #4638, opened 2021-12-01
Make Express and Promt memory configurable

jhonatanamado:
Add an option to configure the max memory per core for Express and Promt configurable via Configfile.

### DIFF
--- a/src/python/T0/RunConfig/RunConfigAPI.py
+++ b/src/python/T0/RunConfig/RunConfigAPI.py
@@ -616,18 +616,10 @@ def configureRunStream(tier0Config, run, stream, specDirectory, dqmUploadProxy):
             specArguments['TimePerEvent'] = streamConfig.Express.TimePerEvent
             specArguments['SizePerEvent'] = streamConfig.Express.SizePerEvent
 
-            if streamConfig.Express.Scenario == "HeavyIonsRun2":
-                baseMemory = 3000
-                perCoreMemory = 1300
-            else:
-                baseMemory = 3500
-                perCoreMemory = 1050
-
-            specArguments['Memory'] = baseMemory + perCoreMemory
-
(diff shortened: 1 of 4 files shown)


src/python/T0/RunConfig/RunConfigAPI.py
```diff
@@ -616,19 +616,13 @@ def configureRunStream(tier0Config, run, stream, specDirectory, dqmUploadProxy):
...
-            if streamConfig.Express.Scenario == "HeavyIonsRun2":
-                baseMemory = 3000
-                perCoreMemory = 1300
-            else:
-                baseMemory = 3500
-                perCoreMemory = 1050
-
-            specArguments['Memory'] = baseMemory + perCoreMemory
-
+            baseMemory = 500
```
germanfgv, line 619: From Dima's comment on Slack, I understand that we don't want to assign this 500MB base memory. We want to assign 2GB/core, and use additianal memory only in case of emergency.

germanfgv (commented):
Let's clarify the `BaseMemory` value. Other than that this seems fine.

src/python/T0/RunConfig/RunConfigAPI.py
jhonatanamado: As mentioned during the meeting we should fit CMS infrastructure requirements that means 2GB/core for any type of job. With the previous formula Tier0 was given in some cases more than 2GB/core so I will start  to test this PR with the runs from 2021 and start to validate which jobs were requiring more than 2GB/core

jhonatanamado:
Changes were tested with different replays. No issues to report only 1 paused job requesting more than 2Gb/core while testing with the run 347028. Same job reported here https://hypernews.cern.ch/HyperNews/CMS/get/tier0-Ops/2321.html

germanfgv (approved):
Everything seems fine! Thanks for the changes @jhonatanamado

PR #4642, opened 2022-01-20
Add lifetime to dataset from ConfigFile

jhonatanamado:
Add lifetime for dataset (RucioContainerRules) that will be subscribed to only disk sites via our ConfigFile

### DIFF
--- a/src/python/T0/RunConfig/RunConfigAPI.py
+++ b/src/python/T0/RunConfig/RunConfigAPI.py
@@ -354,7 +354,8 @@ def configureRunStream(tier0Config, run, stream, specDirectory, dqmUploadProxy):
                                             'autoApproveSites' : autoApproveSites,
                                             'priority' : "high",
                                             'primaryDataset' : specialDataset,
-                                            'deleteFromSource' : True } )
+                                            'deleteFromSource' : True,
+                                            'datasetLifetime' : streamConfig.Express.datasetLifetime } )
 
             alcaSkim = None
             if len(streamConfig.Express.AlcaSkims) > 0:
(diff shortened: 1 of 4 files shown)


src/python/T0/RunConfig/RunConfigAPI.py
```diff
@@ -689,11 +693,13 @@ def configureRunStream(tier0Config, run, stream, specDirectory, dqmUploadProxy):
             factory = RepackWorkloadFactory()
             wmSpec = factory.factoryWorkloadConstruction(workflowName, specArguments)
             for subscription in subscriptions:
+                logging.info("subscrition_bulk %s", subscription)
```
germanfgv, line 696: Typo

src/python/T0/RunConfig/RunConfigAPI.py
```diff
@@ -689,11 +693,13 @@ def configureRunStream(tier0Config, run, stream, specDirectory, dqmUploadProxy):
...
                 wmSpec.setSubscriptionInformation(**subscription)
         elif streamConfig.ProcessingStyle == "Express":
             factory = ExpressWorkloadFactory()
             wmSpec = factory.factoryWorkloadConstruction(workflowName, specArguments)
             for subscription in subscriptions:
+                logging.info("subscrition_express %s", subscription)
```
germanfgv, line 702: Also typo. I'd also review the need for this log entry, and if you consider it to be necessary, then please give it a clearer format so anyone can understand what information is being given.

germanfgv (commented):
Looks good, just a couple of comments about logging

src/python/T0/RunConfig/RunConfigAPI.py
jhonatanamado: This was only for debugging purposes. I updated this PR

src/python/T0/RunConfig/RunConfigAPI.py
jhonatanamado: Same as above

germanfgv (approved):
This is ready to be merged

germanfgv:
@jeyserma Can you comment on the default lifetime that we should give to our subscriptions?

drkovalskyi:
Let's set 3 months as default till we have a better number.

PR #4643, opened 2022-02-10
Add parameters to block-level rules

jhonatanamado:
Add useful parameters for block-level rules via WMAgentConfig.
This changes will be useful to give a lifetime to all the block-level rules that each replay generates. Only small changes on our deployments scripts.
Linked with https://github.com/dmwm/WMCore/pull/10989


jhonatanamado:
@germanfgv , German now that https://github.com/dmwm/WMCore/pull/10989 was merged, this changes needs to be taken into account for future deployments of the T0 agent.

germanfgv (approved):

PR #4649, opened 2022-02-28
Configuration for CRAFT 2022

germanfgv:
Production and Replay configuration used as initial configuration for CRAFT 2022.


tvami:
@germanfgv is something wrong with the syntax checker? How can these changes lead to a syntax check failure?

PR #4654, opened 2022-03-23
Testing `@ctpps` DQM sequence on 2018 ZeroBias dataset

francescobrivio:
# Replay Request

**Requestor**  
Francesco Brivio for PPS/DQM

**Describe the configuration**  
* Release: CMSSW_12_2_1_patch2
* Run: 324841
* GTs:
   * expressGlobalTag: 122X_dataRun3_Express_TIER0_REPLAY_Run2_v1
   * promptrecoGlobalTag: 122X_dataRun3_Prompt_TIER0_REPLAY_Run2_v1
   * alcap0GlobalTag: 122X_dataRun3_Prompt_TIER0_REPLAY_Run2_v1
* Additional changes:
   * Removed all the PDs except for ZeroBias (to save computing resources)
   * Ignored all the express streams (to save computing resources)
   * Set correct eras for running on 2018 data
   * Added `"@ctpps"` DQM sequence to ZeroBias dataset

**Purpose of the test**  
This replay is to test the `"@ctpps"` DQM sequence and conclude #4588 that was never followed up from last year.
I used the configuration from the PPS replay on 2018 data (#4640) which was successful, and only added the new DQM sequence. @germanfgv @jhonatanamado please check the whole config and especially if the `ignoreStream`s are correctly used.

FYI @vavati @fabferro @jfernan2 @tvami @malbouis 

**T0 Operations cmsTalk thread**  
N/A

PR #4655, opened 2022-03-25
Add `@ctpps` dqm sequence on ZeroBias dataset

jhonatanamado:
Add a new dqm sequence for ZeroBias after the test  [#4654](https://github.com/dmwm/T0/pull/4654)


jhonatanamado:
Also tested with another replay and finished with no issues. Ready to be merged.

PR #4650, opened 2022-03-11
Replay to test new Run-3 LumiPCC PCL wf

tvami:
# Replay Request

**Requestor**  

AlCaDB

**Describe the configuration**  
* Release: ~~CMSSW_12_3_0_pre6~~ then ~~CMSSW_12_4_0_pre2~~ then CMSSW_12_3_0
* Run: 346512 (2021 pp)
* GTs:
   * expressGlobalTag: 123X_dataRun3_Express_TIER0_REPLAY_Run2_v1
   * promptrecoGlobalTag: ~123X_dataRun3_Prompt_TIER0_REPLAY_Run2_v2~ then 123X_dataRun3_Prompt_TIER0_REPLAY_Run2_v3
   * alcap0GlobalTag: ~123X_dataRun3_Prompt_TIER0_REPLAY_Run2_v2~ 123X_dataRun3_Prompt_TIER0_REPLAY_Run2_v3
* Additional changes: Commented out old wf-s, added the new AlCa producers 

**Purpose of the test**  

To test new Run-3 LumiPCC PCL wf

https://cms-conddb.cern.ch/cmsDbBrowser/diff/Prod/gts/123X_dataRun3_Prompt_TIER0_REPLAY_Run2_v2/123X_dataRun3_Prompt_TIER0_REPLAY_Run2_v3

Adding the tag from Prep DB produced by a prev replay to test prompt. 

**T0 Operations cmsTalk thread**  

https://cms-talk.web.cern.ch/t/replay-new-lumipcc-pcl-workflows-for-run-3/8060

PR #4668, opened 2022-04-29
Enable prompt AlCaLumiPixelsCountsPrompt

jhonatanamado:
add alca_producers and enable prompt for this PD and change alcaLumiPixelsScenario for Run3

### DIFF
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -108,7 +108,7 @@ hcalnzsScenario = "hcalnzsEra_Run3"
 hiScenario = "ppEra_Run3"
 alcaTrackingOnlyScenario = "trackingOnlyEra_Run3"
 alcaTestEnableScenario = "AlCaTestEnable"
-alcaLumiPixelsScenario = "AlCaLumiPixels"
+alcaLumiPixelsScenario = "AlCaLumiPixels_Run3"
 hiTestppScenario = "ppEra_Run3"
 
 # Defaults for processing version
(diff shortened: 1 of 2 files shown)


jhonatanamado:
GTs needs to be validated and check backward compatibility with the previous `alcaLumiPixelsScenario`

tvami:
> GTs needs to be validated and check backward compatibility with the previous `alcaLumiPixelsScenario`

I'm not really sure what this means. The new scenario `AlCaLumiPixels_Run3` is for Run-3 only, while `AlCaLumiPixels` is for Run-2. There is a new stream for Run-3 as well, so this is by definition not backward compatible.

tvami:
I should also say that the prerequisite of this prompt to work properly is to change the express workflow, as done in https://github.com/dmwm/T0/pull/4650 here:
https://github.com/tvami/T0/blob/cfa2efad806fcade9c100ef7074c2fe2e1da663a/etc/ReplayOfflineConfiguration.py#L230-L253

tvami:
> Are all the streamers and PDs needed for Run3?

Depends on the fill type, for special fills (Van der Meer scans) that happens 1-2 times a year, all the numbers are needed, while in default fills we only have one stream. @jbenitez could comment more

benitezj:
> > Are all the streamers and PDs needed for Run3?
> 
> Depends on the fill type, for special fills (Van der Meer scans) that happens 1-2 times a year, all the numbers are needed, while in default fills we only have one stream. @benitezj could comment more

hi all.
 I think to avoid missing any data lets process both streams (Prompt and Express) for ALL fills.  

Thanks.
Jose

(it is true we don't need the ALCARECO's of the vdM fills, but doesn't hurt to have them and maybe we can use them for some check.)

tvami:
Hi @benitezj maybe I misunderstood @jhonatanamado , but I thought his question was about the fact that there are a lot of numbers after `AlCaLumiPixelsCountsPrompt`
Screen Shot 2022-05-03 at 08 52 42
 
My understanding was that these numbers only come in for the vdM fills, is that correct?

germanfgv:
these changes are being included in #4677

PR #4670, opened 2022-05-06
Update deployment scripts

jhonatanamado:
Update prod/replay deployment scripts to use T0 3.0.4 release


germanfgv (changes requested):
Can you also update the patches file? I think right know we dont need any.

PR #4671, opened 2022-05-11
Decrease priority for Replays

jhonatanamado:
Decrease the basepriority for replays to not interfere with production jobs. Without this changes the priority for the T0 jobs are as follows:
```
1-ExpressReplay (251000 + 10000) = 261k
2-ExpressProd (250000 + 10000) = 260k
3-RepackReplay (251000+ 5000) = 256k
4-RepackProd (250000 + 5000) = 255k
5-PromptReplay = 251k
6-PromptProd = 250k
```
With this changes the priority will be
```
1-ExpressProd (251000 + 10000) = 261k
2-RepackProd (251000 + 5000) = 256k
3-PromptProd = 251k
4-ExpressReplay (240000 + 10000) = 250k
5-RepackReplay (240000 + 5000) = 245k
6-PromptReplay = 240k
```

### DIFF
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -56,7 +56,7 @@ addSiteConfig(tier0Config, "T0_CH_CERN_Disk",
 #  Processing site (where jobs run)
 #  PhEDEx locations
 setAcquisitionEra(tier0Config, "Commissioning2022")
-setBaseRequestPriority(tier0Config, 250000)
+setBaseRequestPriority(tier0Config, 251000)
 setBackfill(tier0Config, None)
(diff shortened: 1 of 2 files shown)


davidlange6:
when  in steady state production (assuming a successful run in 2022), this means no replays... why has this is a problem with comics? The t0 is very small now?

germanfgv:
Its been a problem when running multiple ready at the same time. As you can see, the base priority can be easily changed in our config file, so this is just meant to affect the default behavior of replays. Right now, production jobs are not many and we frequently have several non-critical replays running, so we want the default to prioritize production. It's possible that, in a couple of months, we shall change this back. We will see how it goes.

PR #4687, opened 2022-06-09
Inject Run2022A collisions runs

jhonatanamado:
Changes are according to what is displayed in t0wmadatasvc for each run. This manually injection is due the following issue described in this [Jira](https://its.cern.ch/jira/browse/CMSTZ-990).
@jeyserma @germanfgv Please check the changes. An extra pair of eyes could help if Im missing something.


germanfgv (approved):
It seem correct to me.

jhonatanamado:
vocms013 was used to this operation. Old files were invalidated and erased from rucio and the new files were injected without issues.

PR #4706, opened 2022-07-04
Changing era to Run2022B

jhonatanamado:
Set `raw_to_disk=False` for everything, but HLTPhysics*, EphemeralHLTPhysics* and ZeroBias*

### DIFF
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -55,7 +55,7 @@ addSiteConfig(tier0Config, "T0_CH_CERN_Disk",
 #  Data type
 #  Processing site (where jobs run)
 #  PhEDEx locations
-setAcquisitionEra(tier0Config, "Run2022A")
+setAcquisitionEra(tier0Config, "Run2022B")
 setBaseRequestPriority(tier0Config, 251000)
 setBackfill(tier0Config, None)
 setBulkDataType(tier0Config, "data")
@@ -487,7 +487,6 @@ for dataset in DATASETS:
     addDataset(tier0Config, dataset,
                do_reco=True,
                write_reco=False,
-               raw_to_disk=True,
                write_dqm=True,
                dqm_sequences=["@common"],
                tape_node="T1_US_FNAL_MSS",
@@ -516,7 +515,6 @@ DATASETS = ["DisplacedJet"]
 for dataset in DATASETS:
     addDataset(tier0Config, dataset,
                do_reco=True,
-               raw_to_disk=True,
                write_dqm=True,
                dqm_sequences=["@common"],
                tape_node="T1_FR_CCIN2P3_MSS",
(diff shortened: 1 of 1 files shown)


francescobrivio:
Hi @jhonatanamado!
Maybe I misunderstood Dima's message, but in https://cms-talk.web.cern.ch/t/raw-on-disk-in-run2022b-and-later/12317 I read:

> The only exception that we are aware of is the TSG PDs, i.e. HLTPhysics and ZeroBias PDs in all their different forms, which will be subscribed by default to CERN for 3 months.

Which means that RAW (and all other forms) for HLTPhysics and ZeroBias PDs should stay on Disk for 3 months, am I missing something?
Thanks a lot!

drkovalskyi:
Thanks Francesco. @germanfgv please review the pull request. There is no need to set raw_to_disk to False for each call of addDataset. We set raw_to_disk to False by default. So it should be enough to remove all unnecessary `raw_to_disk=True` flags.

drkovalskyi:
The change of the configuration in the repo doesn't have direct effect on the production node. Beside we will need to switch the machine for new era. Basically it's a preparation for the era change.

etc/ProdOfflineConfiguration.py
```diff
@@ -223,7 +223,7 @@
            archival_node="T0_CH_CERN_MSS",
            tape_node="T1_US_FNAL_MSS",
            disk_node="T1_US_FNAL_Disk",
-           raw_to_disk=False,
+           raw_to_disk=True,
```
germanfgv, line 226: We should not change the config for the default dataset.
```suggestion
           raw_to_disk=False,
```
otherwise, we would need to add `raw_to_disk=False` for every dataset. Makes more sense to keep it False and change the few we need in disk

germanfgv: Also, we don't want undefined datasets (like those that come from minidaq runs) to go to disk by default.

etc/ProdOfflineConfiguration.py
```diff
@@ -973,7 +973,7 @@
 for dataset in DATASETS:
     addDataset(tier0Config, dataset,
                do_reco=True,
-               raw_to_disk=True,
+               raw_to_disk=False,
```
germanfgv, line 976: I think the purpose of this PR is the opposite. We want to send HLTPhysics and ZeroBias RAW to disk, while removing  Raw to disk subscriptions for every other dataset.
https://cms-talk.web.cern.ch/t/raw-on-disk-in-run2022b-and-later/12317

germanfgv (commented):
@jhonatanamado  What we want from the PR is to set `raw_to_disk=False` for all  PRs, with the exception of HLTPhysics and ZeroBias PDs.

Simply keep the default to false, and remove all the unneeded  `raw_to_disk=True` and it should be fine

francescobrivio:
FYI @ttedeschi as this week's ORM :)

jhonatanamado:
Thanks guys, I misunderstand the message. @germanfgv German, could you check again?

germanfgv (approved):
Everything looks fine now. We will prepare a machine with this configuration and make the change to it tomorrow morning, as discussed in https://cms-talk.web.cern.ch/t/era-change/12416

PR #4710, opened 2022-07-09
Test only express for HLTMonitor

jhonatanamado:


jhonatanamado:
Tested and changes merged with https://github.com/dmwm/T0/pull/4711

PR #4715, opened 2022-07-14
Increase TaskArchiver delay to 24 hours (replays)

jeyserma:
Increase TaskArchiver delay for replays from 1 to 24 hours, to keep the completed WFs in the database for further checking/diagnosis. This need to be increased after the reduction of the closing block delay.

### DIFF
--- a/bin/00_deploy_replay.sh
+++ b/bin/00_deploy_replay.sh
@@ -166,7 +166,6 @@ echo 'config.TaskArchiver.logLevel = "DEBUG"' >> ./config/tier0/config.py
 # Workflow archive delay
 #
 #echo 'config.TaskArchiver.archiveDelayHours = 1' >> ./config/tier0/config.py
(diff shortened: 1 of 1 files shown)


bin/00_deploy_replay.sh
```diff
@@ -166,7 +166,7 @@ echo 'config.TaskArchiver.logLevel = "DEBUG"' >> ./config/tier0/config.py
 # Workflow archive delay
 #
 #echo 'config.TaskArchiver.archiveDelayHours = 1' >> ./config/tier0/config.py
-sed -i "s+config.TaskArchiver.archiveDelayHours = 24+config.TaskArchiver.archiveDelayHours = 1+" ./config/tier0/config.py
+sed -i "s+config.TaskArchiver.archiveDelayHours = 24+config.TaskArchiver.archiveDelayHours = 24+" ./config/tier0/config.py
```
germanfgv, line 169: if 24h is the default, then we just need to remove this line and that's it.

PR #4717, opened 2022-07-15
Update to CMSSW_12_4_3 and New Era

jhonatanamado:
Changes taken from https://github.com/dmwm/T0/pull/4716

### DIFF
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -55,7 +55,7 @@ addSiteConfig(tier0Config, "T0_CH_CERN_Disk",
 #  Data type
 #  Processing site (where jobs run)
 #  PhEDEx locations
-setAcquisitionEra(tier0Config, "Run2022B")
+setAcquisitionEra(tier0Config, "Run2022C")
 setBaseRequestPriority(tier0Config, 251000)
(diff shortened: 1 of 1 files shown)


etc/ProdOfflineConfiguration.py
```diff
@@ -1176,6 +1166,22 @@
 ### Parking and Scouting PDs                         ###
 ########################################################
 
+DATASETS = ["ScoutingPFRun3"]
+for dataset in DATASETS:
+    addDataset(tier0Config, dataset,
+               do_reco=True,
+               scenario=ppScenario)
```
francescobrivio, line 1173: in the gDoc I see:
> ScoutingPFRun3 -> scouting = no reco (ScoutingPFMonitor should go in the usual reco)
(about parking see PPD [presentation](https://indico.cern.ch/event/1179710/#41-b-parking-trigger-developme)) (it should be repacked, no reconstruction  and double copy - which is the default behaviour)

So maybe do_reco should be false?

jhonatanamado: Yes my mistake. Its fixed now

germanfgv (approved):
Now it looks good to me. Let's see what the others say

Issue #4526, opened 2020-04-30
Update the T0 deployment script with the Rucio project metadata value


amaltaro:
As recently discussed with the Rucio team, there was a need to distinguish between production and relval NANOAOD containers. In the end, we decided to add a `project` metadata to all blocks/containers inserted by WMAgent.

The GH WMCore issue can be seen here: https://github.com/dmwm/WMCore/issues/9655
And here is a PR implementing it: https://github.com/dmwm/WMCore/pull/9667

As one can see, the default setting is:
```
config.RucioInjector.metaDIDProject = "Production"
```

and what I would like to ask in this ticket, is to adapt this configuration as follows:
* for T0 production activities: change it to `Tier0` in the configuration
* for replays and any other testing activities: change it to `Test` in the config.

germanfgv:
Addressed in #4723

PR #4723, opened 2022-07-25
Update Rucio metadata for T0 prod

germanfgv:
Addresses #4526 
Better late than ever, @amaltaro ?
Let me know is there is something else that needs changing.


amaltaro (approved):
It looks good to me, thanks German.

PR #4724, opened 2022-07-25
Update repack job core splitting

jeyserma:
In the calculation of the #cores for repack jobs, remove the space reservation for streamer files in the scratch area. The total scratch area of 20 GB/core has not changed.


germanfgv:
I'll add this to our patches for 3.0.5

PR #4727, opened 2022-07-27
Enable prompt for ParkingDoubleElectronLowMass* and Update ReplayConfig

jhonatanamado:
Enable prompt for ParkingDoubleElectronLowMass*

### DIFF
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -1107,7 +1107,7 @@ DATASETS = ["ParkingDoubleElectronLowMass0","ParkingDoubleElectronLowMass1","Par
 
 for dataset in DATASETS:
     addDataset(tier0Config, dataset,
-               do_reco=False,
+               do_reco=True,
                tape_node=None,
                scenario=ppScenario)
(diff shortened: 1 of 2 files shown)


jhonatanamado:
From https://hlt-config-editor-confdbv3.app.cern.ch/open?cfg=%2Fcdaq%2Ftest%2Ftomei%2FRun2022%2Fcollisions%2Fstormv1p3%2FHLT%2FV26&db=online, I believe we are correct. `ParkingDoubleElectronLowMass[0-5]`

jhonatanamado:
@jeyserma, Jan Could you please doble-check the replay config.

germanfgv (approved):
It looks fine for me. Some of the changes I don't think are necessary, but also they are not harmful so it's ok. Thank you for re-synching with production config.

jeyserma:
I went through the changes and they look fine to me, nothing to add in particular. It's quite some effort to keep it in sync, so we better find a more suitable solution (that's on me - https://its.cern.ch/jira/projects/CMSTZ/issues/CMSTZ-928)

PR #4729, opened 2022-07-28
Add raise conditions when run/stream can't be configured

jeyserma:
Added 2 raise conditions when a run and/or stream can't be configured. Such errors are severe and should be spotted by the T0 team. These raise conditions will shut down the T0Feeder component.

It follows from the discussion here: https://its.cern.ch/jira/projects/CMSTZ/issues/CMSTZ-995.


germanfgv:
test syntax please

germanfgv:
I think this works well enough for us. Can you add it to the 00_patches.sh script?

PR #4736, opened 2022-08-08
Whitelist T2_CH_CERN_HLT for more Parking* PDs

germanfgv:
We will process additional datasets in HLT in order to further balance the load in our available resources.


germanfgv:
test syntax please

PR #4739, opened 2022-08-09
Replay test of CMSSW_12_4_6

tyjyang:
# Replay Request

**Requestor**  
ORM

**Describe the configuration**  
* Release: CMSSW_12_4_6
* Run: 356824
* GTs:
   * expressGlobalTag: 124X_dataRun3_Express_v4
   * promptrecoGlobalTag: 124X_dataRun3_Prompt_v4
   * alcap0GlobalTag: 124X_dataRun3_Prompt_v4
* Additional changes:

**Purpose of the test**  
This release contains the [fix](https://github.com/cms-sw/cmssw/pull/38989) to be deployed on HLT.

**T0 Operations cmsTalk thread**

PR #4732, opened 2022-08-01
Replay for PPS Sampic PCL workflow and Prompt ALCARECO

francescobrivio:
# Replay Request

**Requestor**  
Francesco Brivio for AlCaDB/PPS

**Describe the configuration**  
* Release: ~CMSSW_12_4_4~ ~CMSSW_12_4_5~ CMSSW_12_4_6
* Run: 356433
* GTs:
   * expressGlobalTag: ~124X_dataRun3_Express_v4~ 124X_dataRun3_Express_Candidate_2022_08_05_17_14_24
   * promptrecoGlobalTag: 124X_dataRun3_Prompt_v4
   * alcap0GlobalTag: 124X_dataRun3_Prompt_v4
* Additional changes:
   * Set the `alcaPPSScenario` to be `"AlCaPPS_Run3"`
   * Added the PPS PCL workflow `PromptCalibProdPPSDiamondSampic`
   * Set the options `do_reco=True, write_reco=False, write_aod=False, write_miniaod=False` for the `AlCaPPSPrompt` dataset in order to produce as output only RAW and the `PPSCalMaxTracks` ALCARECO at Prompt level
   * Ignored all streams except `ALCAPPSExpress` and `ALCAPPSPrompt`

**Purpose of the test**  
Now that with the hlt menu v1.3 we have two different PPS streams for Express and Prompt we want to test the full PPS chain, including PCL and RAW/Prompt-reco.

**T0 Operations cmsTalk thread**  
N/A

FYI @malbouis @tvami @saumyaphor4252 @vavati

PR #4740, opened 2022-08-10
Change CMSSW version to 12_4_6

germanfgv:
Update Production and Replay configuration to CMSSW_12_4_6 after successful replay in #4739
Add Sampic alca produce after successful replay in #4732


tvami:
GT should be `124X_dataRun3_Express_v5`

francescobrivio:
@germanfgv if you mean to integrate #4732 here as well I think you missed a few points:
 - alcaPPSScenario = "AlCaPPS_Run3"
 - for the `AlCaPPSPrompt` Dataset you missed the update of the parameters: `do_reco=True, write_reco=False, write_aod=False, write_miniaod=False,`

PR #4741, opened 2022-08-13
Run2022D and data rebalancing

jhonatanamado:


drkovalskyi:
EphemeralHLTPhysics needs to place RAW on Disk at CERN for 3 months.

drkovalskyi:
In general, EphemeralHLTPhysics* should follow EphemeralZeroBias* configuration, i.e. it writes RAW to disk only. One way or another, we must keep it on disk.

germanfgv:
> In general, EphemeralHLTPhysics* should follow EphemeralZeroBias* configuration, i.e. it writes RAW to disk only. One way or another, we must keep it on disk.

I made the changes and now Ephemeral have archival subscriptions to T0 Tape and also RAW data to T2 CERN Disk

PR #4748, opened 2022-08-25
Replay test for 12_4_8

sarafiorendi:
# Replay Request

**Requestor**  
ORM

**Describe the configuration**  
* Release: CMSSW_12_4_8
* Run: [356824](https://cmsoms.cern.ch/cms/runs/report?cms_run=356824), [356948](https://cmsoms.cern.ch/cms/runs/report?cms_run=356948) (same as for the [CMSSW_12_4_7 replay](https://github.com/dmwm/T0/pull/4745))
* GTs:
   * expressGlobalTag: 124X_dataRun3_Express_v5
   * promptrecoGlobalTag: 124X_dataRun3_Prompt_v4
   * alcap0GlobalTag: 124X_dataRun3_Prompt_v4
* Additional changes: None

**Purpose of the test**  
Test the new release that will be used for the ReReco of the first part of 2022 data and for data-taking, for which a few fixes are included.

**T0 Operations cmsTalk thread**  
https://cms-talk.web.cern.ch/t/replay-testing-for-cmssw-12-4-8/14321

PR #4750, opened 2022-08-29
Change CMSSW version to 12_4_8

germanfgv:
Update Production and Replay configuration to CMSSW_12_4_8 after successful replay in https://github.com/dmwm/T0/pull/4748


etc/ReplayOfflineConfiguration.py
```diff
@@ -1640,6 +1655,11 @@
 ignoreStream(tier0Config, "streamL1Rates")
 ignoreStream(tier0Config, "streamDQMRates")
 
+ignoreStream(tier0Config, "ALCALumiPixelsCountsExpress")
+ignoreStream(tier0Config, "ALCALumiPixelsCountsPrompt")
+ignoreStream(tier0Config, "Calibration")
+ignoreStream(tier0Config, "Express")
+ignoreStream(tier0Config, "NanoDST")
```
tvami, line 1662: This was specific to the replay I did, we should not forget to remove these lines at the next full replay

PR #4761, opened 2022-09-26
Enable Prompt for Ephemeral datasets

germanfgv:
Added T0 Tape as tape_node instead of archival, as archivals does not work for reco output.


germanfgv:
test syntax please

PR #4765, opened 2022-10-04
Add specifyStreams function

gkfthddk:
Add specifyStreams(config, streamNames) at Tier0Config.py

Add speicifyStreams at end of [ReplayOfflineConfiguration.py](https://github.com/dmwm/T0/blob/master/etc/ReplayOfflineConfiguration.py) to ignore every stream except specified streams.
for example,
specifyStreams(tier0Config, ["Calibration","LookArea"])

Replay at vocms0500

- Run: 356005,359060
- DeploymentID: 221012133000
- Additional changes: `specifyStreams(tier0Config, ["Calibration"])` at the end of [ReplayOfflineConfiguration.py](https://github.com/dmwm/T0/pull/4765/files#diff-4d2c96d3e15eb107f3aec15b5259355379fad0c475f704ed3a3cf8720cf94684)

### DIFF
--- a/src/python/T0/RunConfig/Tier0Config.py
+++ b/src/python/T0/RunConfig/Tier0Config.py
@@ -24,6 +24,8 @@ Tier0Configuration - Global configuration object
 | |       |
 | |       |--> InjectMaxRun - Highest run number to be injected into the system from the SM database (default is unlimited)
 | |       |
+| |       |--> SpecifiedStreamNames - Streamer names to be ran exclusivley(default is None)
+| |       |
 | |       |--> AcquisitionEra - The acquisition era for the run
 | |       |
 | |       |--> Backfill - The backfill mode, can be None, 1 or 2
@@ -258,6 +260,8 @@ def createTier0Config():
     tier0Config.Global.InjectRuns = None
     tier0Config.Global.InjectMinRun = None
     tier0Config.Global.InjectMaxRun = None
+    
+    tier0Config.Global.SpecifiedStreamNames = None
 
     tier0Config.Global.ScramArches = {}
     tier0Config.Global.Backfill = None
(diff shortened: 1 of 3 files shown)


germanfgv:
This wont work as athe streams are addit to the configuration after we pull the data from storage manager

germanfgv:
Here is were we get the data from storage manager. After this, the tier0Config object is populated.
https://github.com/dmwm/T0/blob/c4cc364ce6ae87861e03c833071368a298861466/src/python/T0Component/Tier0Feeder/Tier0FeederPoller.py#L156-L178

You need to set the Processing type as `ignore` after this.

gkfthddk:
You mean even I set ignore to streams of tier0Config at the end of ReplayOfflineConfiguration.py, more streams will be loaded after?
Then, streams added after are also needed to be ignored?

src/python/T0/RunConfig/Tier0Config.py
```diff
@@ -786,6 +790,24 @@ def ignoreStream(config, streamName):
...
+    """
+    config.Global.SpecifiedStreamNames = streamNames
+
+    if(type(streamNames)!=list):
+        streamNames=[streamNames]
+    for stream in config.Streams.dictionary_().keys():
+        if not stream in streamNames:
+            streamConfig = retrieveStreamConfig(config, stream)
+            streamConfig.ProcessingStyle = "Ignore"
+
```
germanfgv, line 808: After your last changes, this part is no longer needed.

src/python/T0/RunConfig/Tier0Config.py
```diff
@@ -24,6 +24,8 @@
 | |       |
 | |       |--> InjectMaxRun - Highest run number to be injected into the system from the SM database (default is unlimited)
 | |       |
+| |       |--> SpecifiedStreamNames - Streamer names to be ran exclusivley(default is None)
```
germanfgv, line 27: Thanks for adding documentation. More than 'default is none', it would be better to show that `None` means that **all** streamers will be processed.

PR #4768, opened 2022-10-11
Replay to test CMSSW_12_4_10

sarafiorendi:
# Replay Request

**Requestor**  
Sara Fiorendi for the ORM

**Describe the configuration**  
* Release: CMSSW_12_4_10  
* Run: 359762  
* GTs (unchanged):
   * expressGlobalTag: 124X_dataRun3_Express_v5  
   * promptrecoGlobalTag: 124X_dataRun3_Prompt_v4  
   * alcap0GlobalTag: 124X_dataRun3_Prompt_v4  
* Additional changes:

**Purpose of the test**  
Test the new CMSSW_12_4_10 release in view of the change in the T0 configuration.

**T0 Operations cmsTalk thread**  
https://cms-talk.web.cern.ch/t/asking-replay-for-cmssw-12-4-10/16171

PR #4769, opened 2022-10-12
Update production node to use CMSSW_12_4_10

jhonatanamado:


jhonatanamado:
> what about the replay config?

hi @francescobrivio .To update that config file, I will merge https://github.com/dmwm/T0/pull/4768

PR #4772, opened 2022-10-26
ProdConfig file to reprocess Run_359750

jhonatanamado:
**THIS PR MUST NOT BE MERGED, IS JUST TO HAVE SOME RECORD ABOUT THIS REPROCESSING**
The following run from Run2022E is not completely repacked. https://dmytro.web.cern.ch/dmytro/cmsprodmon/tier0.php?run=359750
Thanks to @germanfgv and SM, they found that this particular run closed before sending the final set of files, likely as a consequence of some problems SM experience with some large files https://cms-talk.web.cern.ch/t/large-streamer-files-in-run-359750/15795.
The course of action was the following:

1. The run was closed.
2. All the T0 workflows were closed
3. Latter T0 received more files

As a consequence 172 files were never repacked. If T0 needs to reprocess that Run, we must discuss if:
- Should we only do the Repack, or should we reprocess all the wfs (Repack,Express,Reco). 

Base configuration for this file was taken from https://github.com/dmwm/T0/pull/4759 

FYI @drkovalskyi @jeyserma @germanfgv


jeyserma:
Thanks, Jhonatan, I have two comments:
- we should ignore the Express streams
- set the promptReco delay to 1 hour (instead of the default value of 2 days)

jhonatanamado:
Thanks @jeyserma . The files that will be invalidated are in `/afs/cern.ch/user/c/cmst0/public/Files2Invalidate/359750_invalidationList.txt`.
Please double-check that list.

jeyserma:
Looks good to me. Shall we still reduce the PromptReco delay to 1 hour?

germanfgv:
Closing this issue as the recovery was successfull and duplicate files have been invalidated

PR #4783, opened 2022-11-19
Change acquisition era to Run2022G

germanfgv:
Change era to Run2022G after HIRun2022A

Also adds `aod_to_disk` attribute to the configuration


francescobrivio:
@germanfgv are we also going to change to `CMSSW_12_4_11_patch1`?
In that case we should also move to the new GTs (same used also in HI):
 - Express GT: 124X_dataRun3_Express_v9
 - Prompt GT: 124X_dataRun3_Prompt_v10

germanfgv:
@francescobrivio We are not moving to a new configuration yet. We will do so on Monday.

PR #4787, opened 2022-11-22
Set T2_CH_CERN_HLT in whitelist by default

germanfgv:
Set T2_CH_CERN_HLT in whitelist by default in all our PromptReco workflows.

This will help keep our resource usage balanced, as all jobs can be picked up by both sides


germanfgv:
@jeyserma @jhonatanamado What do you think?

jeyserma:
No comments; agreed to add HLT to the whitelist for promptreco

jhonatanamado:
Looks good.

PR #4789, opened 2023-02-03
Adding SITECONFIG_PATH var to WN environment

germanfgv:
Fix to the issues with CMSSW override catalog reported here:
https://cms-talk.web.cern.ch/t/invalid-override-string-overriding-tfc-in-cmssw-12-6-2/19904/5


germanfgv:
This is no longer needed after WMCore patches #11473 and #11481

PR #4795, opened 2023-03-07
start branch for 13_0_0 replay

davidlange6:
# Replay Request

**Requestor**  
ORM

**Describe the configuration**  
* Release:  CMSSW_13_0_0
* Run: 364158, 350966, 359691
* GTs:
   * expressGlobalTag:  130X_dataRun3_Express_Candidate_2023_03_07_09_53_07
   * promptrecoGlobalTag: 130X_dataRun3_Prompt_Candidate_2023_03_09_09_47_16
* Additional changes: still need to incorporate an alca matrix update

**Purpose of the test**  
First test of CMSSW_13_0_0

PR #4798, opened 2023-03-14
Production configuration for CRAFT and new processing version

germanfgv:
Production configuration for CRAFT and CMSSW 13_X release cycle.
Includes:

- Processing version moving from 1 to 2 (due to major CMSSW version change)
- Splitting of EGamma, JetMET, and Muon datasets
- Inclusion of "AlCaPCCRandom" for "AlCaLumiPixelsCountsPrompt0"

This configuration was tested in #4795

### DIFF
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -94,11 +94,11 @@ setPromptCalibrationConfig(tier0Config,
 
 # Defaults for CMSSW version
 defaultCMSSWVersion = {
-    'default': "CMSSW_12_6_4"
+    'default': "CMSSW_13_0_0"
 }
 
 # Configure ScramArch
-setDefaultScramArch(tier0Config, "el8_amd64_gcc10")
+setDefaultScramArch(tier0Config, "el8_amd64_gcc11")
 
 # Configure scenarios
 ppScenario = "ppEra_Run3"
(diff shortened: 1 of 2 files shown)


etc/ProdOfflineConfiguration.py
```diff
@@ -113,23 +113,22 @@
...
 
 expressProcVersion = {
-    'default': "1"
+    'default': "2"
 }
 
 # Defaults for GlobalTag
-expressGlobalTag = "126X_dataRun3_Express_v3"
-promptrecoGlobalTag = "126X_dataRun3_Prompt_v3"
-alcap0GlobalTag = "126X_dataRun3_Prompt_v3"
```
francescobrivio, line 132: I think you deleted `alcap0GlobalTag` by mistake

fabiocos, line 132: @germanfgv is this dropped on purpose?

etc/ReplayOfflineConfiguration.py
```diff
@@ -126,9 +130,8 @@
 alcarawProcVersion = dt
 
 # Defaults for GlobalTag
-expressGlobalTag = "126X_dataRun3_Express_v3"
-promptrecoGlobalTag = "126X_dataRun3_Prompt_v3"
-alcap0GlobalTag = "126X_dataRun3_Prompt_v3"
```
fabiocos, line 131: same question here

etc/ProdOfflineConfiguration.py
tvami: but where is `alcap0GlobalTag` used anyway?

germanfgv: it was not by mistake

germanfgv: Yes, it was on porpuse. `alcap0GlobalTag` is not being used anymore (has not been used for years, really).

fabiocos: ok, thanks

francescobrivio: yes indeed you are right @germanfgv , by searching in github this only appears in the Run1 configuration 😄 sorry for the noise!

francescobrivio:
@germanfgv should this PR be merged since it's already deployed in production?

PR #4806, opened 2023-04-20
[RFC] Update ParkingDoubleElectronLowMass PD name

francescobrivio:
## Config Update Request
Following [this comment](https://its.cern.ch/jira/browse/CMSHLT-2599?focusedCommentId=4802897&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4802897) from @missirol regarding the preparations of the new collision menu at 13.6 TeV (expected tomorrow): for 2023 collisions the `ParkingDoubleElectronLowMass[0-5]` have been merged and renamed to `ParkingDoubleElectronLowMass`.

In this PR I'm simply adding the new dataset name (in both prod and replay configs) so that it can be Prompt-reconstructed correctly.

**Important Note:**
In the prod configuration there are actually 2 blocks to handle `ParkingDoubleElectronLowMass[0-5]` datasets:
1. the first block contains only the "0" dataset
2. the second block contains "1-5" datasets

The only difference of the two blocks is that the second one has `aod_to_disk=False`.
I added the the new dataset to the _first_ one: this means AOD will be sent to disk (IIUC) --> **Please confirm if this is ok or not!**

https://cms-talk.web.cern.ch/t/update-parkingdoubleelectronlowmass-pd-name/22966

Attn. @fabiocos @germanfgv @trtomei @mzarucki

### DIFF
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -1090,7 +1090,7 @@ for dataset in DATASETS:
                disk_node="T1_UK_RAL_Disk",
                scenario=ppScenario)
 
-DATASETS = ["ParkingDoubleElectronLowMass0"]
+DATASETS = ["ParkingDoubleElectronLowMass","ParkingDoubleElectronLowMass0"]
 for dataset in DATASETS:
     addDataset(tier0Config, dataset,
                do_reco=True,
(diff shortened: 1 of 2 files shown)


francescobrivio:
test syntax please

fabiocos:
@GermanGiraldo this is just a renaming, and it should go together with the update of Era to be done tomorrow late morning. According to the HLT DOC there should not be any further change.

germanfgv:
We will merge this with #4807

PR #4816, opened 2023-05-09
Add subscription of RAW to disk for most datasets

germanfgv:
Subscribing RAW data to disk to help in future reReco campaigns


germanfgv:
test systax please

germanfgv:
We will not move on with this, as the amount of data sent to disk sites (specially FANL) would be to high

PR #4827, opened 2023-05-24
Nano aod addition

LinaresToine:
Implement nanoAOD

PR #4830, opened 2023-05-25
Add NanoAOD to T0 configuration

germanfgv:
Adds NanoAOD output to all datasets that produce MiniAOD
PR to be used on top of #4827


germanfgv:
What we use to prevent this is `do_reco=False`. If we are required to reconstruct, then we default to produce all datatiers unless told otherwise.

Your comment makes sense, but I guess it will depend on what we agree on with PPD. For now, we will mimic MiniAOD's config.

germanfgv:
test syntax please

PR #4836, opened 2023-06-07
Include nanoaod in Tier0DataSvc

LinaresToine:
This PR with update in the Tier0FeederPoller.py is required to include the modification in a patch for correct agent deployment.
It also includes changes needed in the GetRecoConfigs and InsertRecoConfigs files


LinaresToine:
Replay is running to test this patch. The changes seem to function correctly in T0AST and T0DataSvc databases.

It may be followed [here](https://monit-grafana.cern.ch/d/t_jr45h7k/cms-tier0-replayid-monitoring?orgId=11&refresh=1m&var-Bin=5m&var-ReplayID=230612151819&var-JobType=All&var-WorkflowType=All)

PR #4849, opened 2023-07-10
Adding secondary dataset for DoubleMuonLowMass

jeyserma:
Adding support to create a secondary dataset for the Parking DoubleMuonLowMass.

The secondary dataset is created in a PromptReco workflow that produces RAW output. The storage of this secondary PD is identical to the main RSEs for the dataset.


germanfgv:
@jeyserma are there any updates regarding this issue?

drkovalskyi:
What are you expecting? I think it's a fundamentally broken approach. We have time to sort it out by Run2024

PR #4850, opened 2023-07-12
Replay Request to test limitation on lumisections

ItyTell:
Requestor
T0

Release:  CMSSW_13_0_9
Run: 368823
GTs:
    expressGlobalTag: 130X_dataRun3_Express_v3
    promptrecoGlobalTag: 130X_dataRun3_Prompt_v4
Additional changes:  adding a new parameter to limit the number of lumisections to inject into the system.

Purpose of the test
An attempt to limit the number of lumisections in the replay test. It will be used in new small functional test in the future.

### DIFF
--- a/etc/ReplayOfflineConfiguration.py
+++ b/etc/ReplayOfflineConfiguration.py
@@ -38,7 +38,8 @@ setConfigVersion(tier0Config, "replace with real version")
 # 367102 - Collisions 2023 - 1200b - 0.5h long - all components IN
 setInjectRuns(tier0Config, [369998])
 
-setInjectLimit(tier0Config, 10)
+# Use this in order to limit the number of lumisections to process
+#setInjectLimit(tier0Config, 10)
 
 # Settings up sites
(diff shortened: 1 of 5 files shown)


germanfgv:
The changes look great! Let's test run a replay with it an see how it goes.

etc/ReplayOfflineConfiguration.py
```diff
@@ -37,6 +38,8 @@
 # 367102 - Collisions 2023 - 1200b - 0.5h long - all components IN
 setInjectRuns(tier0Config, [369998])
 
+setInjectLimit(tier0Config, 10)
```
germanfgv, line 41: Could you please comment out the call? We don't want the limitation to be the default in the master branch:
```suggestion
# Use this in order to limit the number of lumisections to process
# setInjectLimit(tier0Config, 10)
```

germanfgv (commented):
Looks ready. Please address the comment I mande and then we can berge it.

germanfgv:
Thank you @ItyTell

PR #4857, opened 2023-07-28
Add files that support the primary dataset config history API entry

LinaresToine:
A new T0 API entry has been requested by DQM/PPD to keep track of the configuration used for each primary dataset as their respective configuration changes. Such request can be seen in this Jira ticket: https://its.cern.ch/jira/browse/CMSTZDEV-793?filter=-1

A replay will be launched with this changes included.


germanfgv:
Same as in: https://github.com/dmwm/T0/pull/4858#issuecomment-1674467751

can we close this?

PR #4858, opened 2023-07-28
2 api primary dataset config

LinaresToine:
A second alternative for the new T0 API entry described in PR #4857


germanfgv:
@LinaresToine can you close this PR? or is there something else to test with this?

PR #4868, opened 2023-08-29
Tentative config file for T2_US_Vanderbilt storage testing

LinaresToine:
This config file attempts to test T2_US_Vanderbilt site as a storage site. It replays four runs: 359694, 359762, 359764 and 366035 for a total of 242.5 TB. I kindly request @GermanGiraldo to check the naming I've used for T2_US_Vanderbilt, as well as details regarding the maxRunPreviousConfig variable, since run 366035 used another CMSSW version. It would still be important to add the different global tags after the max run.

Thank you for your insight.

### DIFF
--- a/etc/ReplayOfflineConfiguration.py
+++ b/etc/ReplayOfflineConfiguration.py
@@ -35,14 +35,14 @@ setConfigVersion(tier0Config, "replace with real version")
 
 # Set run number to replay
 # 367102 - Collisions 2023 - 1200b - 0.5h long - all components IN
-setInjectRuns(tier0Config, [369998])
+setInjectRuns(tier0Config, [359694, 359762, 359764, 366035])
 
 # Settings up sites
 processingSite = "T2_CH_CERN"
-storageSite = "T0_CH_CERN_Disk"
-streamerPNN = "T0_CH_CERN_Disk"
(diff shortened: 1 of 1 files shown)


LinaresToine:
Run 366035 has been removed from the replay to avoid issues with incompatibilities in GTs. The idea is to run 3 simultaneous replays to simulate the high input data.

etc/ReplayOfflineConfiguration.py
```diff
@@ -182,8 +195,8 @@
            global_tag=promptrecoGlobalTag,
            global_tag_connect=globalTagConnect,
            #archival_node="T0_CH_CERN_MSS",
-           tape_node="T0_CH_CERN_Disk",
-           disk_node="T0_CH_CERN_Disk",
+           #tape_node="T1_US_FNAL_MSS",
+           disk_node="T2_US_Vanderbilt",
            #raw_to_disk=False,
```
germanfgv, line 200: You can put:
```suggestion
           raw_to_disk=True,
```
In the default dataset configuration (and any other PD for that matter)

germanfgv (changes requested):
Looks good to me. Activate RAW to disk so we transfer all output to Vanderbilt and it should be ready

germanfgv:
@LinaresToine can we close this PR?

PR #4863, opened 2023-08-18
HIReplay configuration for testing CMSSW_13_2_1

germanfgv:
# Replay Request

**Requestor**  
RC and PPD

**Describe the configuration**  
* Release: CMSSW_13_2_1
* Run: 362321 (2022 HI run)
* GTs:
   * expressGlobalTag: 132X_dataRun3_Express_v2
   * promptrecoGlobalTag: 132X_dataRun3_Prompt_v2
* Additional changes:
Adds 2 additional scenarios to the `event_scenario` table

**Purpose of the test**  
Test a CMSSW_13_2_X release in preparation for resuming operations at LHC and 2023 HI run

**T0 Operations cmsTalk thread**
https://cms-talk.web.cern.ch/t/testing-of-cmssw-13-2-x-releases-before-restart-of-operations/28475/6?u=ggiraldo

PR #4872, opened 2023-09-14
Remove NanoAOD from RAW' dataset

germanfgv:
RAW' data is not expected to be able to produce NanoAOD, as discussed in #4863. Here we disabling it's production for `CommissioningRAWPrime` PD.


mmusich:
> RAW' data is not expected to be able to produce NanoAOD

just to clarify, it's not the RAW' per se that cannot produce NanoAOD, but rather that the NanoAOD sequence was not tested for Heavy Ion events (see [here](https://mattermost.web.cern.ch/cms-exp/channels/orm--t0/msgqsbazofd7bkjyqiesytyzec)).

PR #4876, opened 2023-09-21
Add T1 sites to siteWhitelist

LinaresToine:


germanfgv:
Adding T1 sites for additional PromptReco power during HI run

PR #4877, opened 2023-09-27
Change scenario in HIForward datasets

LinaresToine:
Change proposed in https://cms-talk.web.cern.ch/t/possible-issue-with-t0-config-for-hi-prompt-reco/29972


etc/HIProdOfflineConfiguration.py
```diff
@@ -1483,7 +1483,7 @@
                                "HcalCalIsolatedBunchSelector", "HcalCalIterativePhiSym","HcalCalMinBias",
                                "TkAlJpsiMuMu", "TkAlUpsilonMuMu","TkAlZMuMu","TkAlMuonIsolated"],
                dqm_sequences=["@commonSiStripZeroBias", "@ecal", "@hcal", "@muon", "@jetmet", "@egamma"],
-               scenario=hiTestppScenario)
+               scenario=ppEra_Run3_2023)
```
mmusich, line 1486: Should not this be `scenario=ppScenario` ?

See
https://github.com/dmwm/T0/blob/6d2579ffd3c86b3449c1bf3b0b44dca7f79e2689/etc/HIProdOfflineConfiguration.py#L105

germanfgv:
We are testing this configuration in #4878

germanfgv:
Given the results in #4882 we will not perform this change in the near future

PR #4881, opened 2023-09-29
Sync HIReplay configuration with HI Production

germanfgv:
This PR updates the HIReplay configuration file so it is in sync with HIProdOfflineconfiguration after #4875


germanfgv:
test syntax please

PR #4899, opened 2023-10-20
Adding unit test for "Tier0Config"

ItyTell:
Adding unit test for "Tier0Config".

### DIFF
--- a/test/python/T0_t/Config/Tier0Config_t.py
+++ b/test/python/T0_t/Config/Tier0Config_t.py
@@ -1,0 +1,555 @@
+import unittest
+import T0.RunConfig.Tier0Config as tier0config
+
+
+class TestRetrieveStreamConfig(unittest.TestCase):
+
+    def setUp(self):
+        self.config = tier0config.createTier0Config()
+
+    def test_createTier0Config(self):
+        # Check some key values in the Global section
+        self.assertEqual(self.config.Global.ProcessingSite, "T2_CH_CERN")
+        self.assertEqual(self.config.Global.StreamerPNN, "T0_CH_CERN_Disk")
+        self.assertEqual(self.config.Global.StorageSite, "T0_CH_CERN_Disk")
+        self.assertEqual(self.config.Global.DQMDataTier, "DQMIO")
+        self.assertEqual(self.config.Global.BaseRequestPriority, 150000)
+        self.assertEqual(self.config.Global.EnableUniqueWorkflowName, False)
+        self.assertEqual(self.config.Global.DeploymentID, 1)
+
+    def test_retrieveStreamConfig(self):
+        # Test with an existing stream
+        streamName = "Express"
+        streamConfig = tier0config.retrieveStreamConfig(self.config, streamName)
+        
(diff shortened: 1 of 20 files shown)


germanfgv:
@ItyTell looks good, could you please clarify how you are running the tests? I want to reproduce your steps.

ItyTell:
@germanfgv just use "python3 Tier0Config_t.py". (python should have WMCore, T0 and unittest packages installed). I tried it in docker container but should work in any system too.

test/python/T0_t/Config/Tier0Config_t.py
```diff
@@ -0,0 +1,557 @@
...
+        streamConfig = tier0config.retrieveStreamConfig(self.config, nonExistentStreamName)
+
+        # Check if the default values were copied correctly
+        self.assertEqual(streamConfig._internal_name, nonExistentStreamName)
+        self.assertEqual(getattr(self.config.Streams, nonExistentStreamName), streamConfig)
+
+    def test_deleteStreamConfig(self):
+        # Test with an existing stream
+        streamName = "Express"
+        tier0config.deleteStreamConfig(self.config, streamName)
```
germanfgv, line 32: AS there was no `Express` stream in the first place, this test would succeed even if the function does not work. We could thest this after testing addRepackConfig, so there is a stream to delete.

test/python/T0_t/Config/Tier0Config_t.py
```diff
@@ -0,0 +1,557 @@
...
+
+        # Check if the stream configuration has been deleted
+        self.assertIsNone(getattr(self.config.Streams, streamName, None))
+
+    def test_retrieveSiteConfig(self):
+        # Test with an existing site
+        siteName = "T1_US_FNAL"
+        siteConfig = tier0config.retrieveSiteConfig(self.config, siteName)
```
germanfgv, line 40: The has not been added, therefore it does not really exist. This is similar to the comments about `test_retrieveStreamConfig`

test/python/T0_t/Config/Tier0Config_t.py
```diff
@@ -0,0 +1,557 @@
...
+        # Test with an existing dataset
+        datasetName = "PrimaryDataset"
+        datasetConfig = tier0config.retrieveDatasetConfig(self.config, datasetName)
+        
+        # Test with a non-existing dataset, it should fall back to "Default"
+        nonExistentDatasetName = "NonExistentDataset"
+        datasetConfig = tier0config.retrieveDatasetConfig(self.config, nonExistentDatasetName)
+
+        # Check if the default values were copied correctly
+        self.assertEqual(datasetConfig._internal_name, nonExistentDatasetName)
```
germanfgv, line 68: Same as `test_retrieveStreamConfig`

germanfgv (commented):
Other than the issues I have with the first group of unit tests, I thing this should work prety well. 

Another detail: Please change the path from `test/python/T0_t/Config/Tier0Config_t.py` to `test/python/T0_t/RunConfig_t/Tier0Config_t.py`

germanfgv (approved):
Much better. Thank you @ItyTell

PR #4902, opened 2023-11-09
deployment scripts with modified job resubmission

LinaresToine:
As described in the jira ticket https://its.cern.ch/jira/browse/CMSTZDEV-800, the automatic job resubmission could be friendlier to operators when processing jobs fail due to temporary issues, as well as SCRAM errors that result in failed jobs after 0 retries. This PR is the interpreted proposal from said jira ticket.


germanfgv:
Looks good. Can we merge this?

PR #4914, opened 2024-02-09
Replay for CMSSW_14_0_0_pre3

germanfgv:
# Replay Request

**Requestor**  
Team or person that requests this replay

**Describe the configuration**  
* Release: CMSSW_14_0_0_pre3
* Run: 369998, 359688, 375832
* GTs:
   * expressGlobalTag: 140X_dataRun3_Express_140XReplay_v1
   * promptrecoGlobalTag: 140X_dataRun3_Prompt_140XReplay_v1
* Additional changes:
   * Express: Add TkAlZMuMu and PromptCalibProdSiPixelAliHGComb
   * HLTPhysics*: Add TkAlV0s producer 

**Purpose of the test**  
First test for the CMSSW_14_X release cicle and new AlCa producers and GTs

**T0 Operations cmsTalk thread**  
[Tier0 Operations cmsTalk Forum](https://cms-talk.web.cern.ch/t/replay-request-with-cmssw-14-0-0-pre3/34380)

PR #4918, opened 2024-02-20
Replay configuration for CMSSW_14_0_0 using #4914 as starting point

fabiocos:
# Replay Request

**Requestor**  
ORM

* Release: CMSSW_14_0_0
* Run: 369998, 359688, 375832
* GTs:
   * expressGlobalTag: 140X_dataRun3_Express_v2
   * promptrecoGlobalTag: 140X_dataRun3_Prompt_v2
* Additional changes:
   * Express: Add TkAlZMuMu and PromptCalibProdSiPixelAliHGComb
   * HLTPhysics*: Add TkAlV0s producer

**Purpose of the test**  
Initial test of CMSSW_14_0_0, based on the configuration used for the earlier replay of 14_0_0_pre3 #4914 , pending full release validation

[Tier0 Operations cmsTalk Forum](https://cms-talk.web.cern.ch/t/replay-request-with-cmssw-14-0-0/35240)

### DIFF
--- a/etc/ReplayOfflineConfiguration.py
+++ b/etc/ReplayOfflineConfiguration.py
@@ -110,10 +110,10 @@ defaultCMSSWVersion = {
 }
 
 # Configure ScramArch
-setDefaultScramArch(tier0Config, "el8_amd64_gcc12")
+setDefaultScramArch(tier0Config, "el8_amd64_gcc11")
 setScramArch(tier0Config, "CMSSW_12_4_9", "el8_amd64_gcc10")
 setScramArch(tier0Config, "CMSSW_12_3_0", "cs8_amd64_gcc10")
-setScramArch(tier0Config, "CMSSW_13_3_2_patch1", "el8_amd64_gcc12")
+setScramArch(tier0Config, "CMSSW_14_0_0", "el8_amd64_gcc12")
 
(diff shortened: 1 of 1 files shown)


germanfgv:
run replay please

fabiocos:
@germanfgv do you understand the reason for the failure? The Jenkins console output does not sound that informative to me

fabiocos:
@germanfgv is the T0 supposed to have el8_amd64_gcc11 as default in next future, and we should take care of this for each possible new version to be tested? I expect a 13_3_2_patch2 very soon for instance

germanfgv:
> @germanfgv is the T0 supposed to have el8_amd64_gcc11 as default in next future, and we should take care of this for each possible new version to be tested? I expect a 13_3_2_patch2 very soon for instance

As it works now, we define a default architecture, and we specify any exceptions that we might need. This is particularly important for a replay like this which is using runs from different years. In Production, it is usually enough to define a default architecture, except in cases in which we are moving from one architecture to the other (e.g. last week).

We will move towards automatically verifying the proper architecture while creating the workflows, to avoid this problems in the future.

germanfgv:
Continuing 14_0_0 tests in #4924

PR #4967, opened 2024-07-08
repay config

lucalavezzo:
# Replay Request

**Requestor**  
ORM

* Release: CMSSW_14_0_10
* Run: 382686, 382726  
* GTs:
   * expressGlobalTag: 140X_dataRun3_Express_v3
   * promptrecoGlobalTag:140X_dataRun3_Prompt_v4
* Additional changes:

**Purpose of the test**  
Test new release, CMSSW_14_0_10

### DIFF
--- a/etc/ReplayOfflineConfiguration.py
+++ b/etc/ReplayOfflineConfiguration.py
@@ -36,10 +36,10 @@ tier0Config = createTier0Config()
 setConfigVersion(tier0Config, "replace with real version")
 
 # Set run number to replay
-# 379070 - Cosmics 3.8T 2024 - 3h long - ALL components in
-# 380128 - 2024 13.6 TeV Collision run - 2340b - 2.5h long - ALL components in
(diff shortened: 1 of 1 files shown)


germanfgv:
Hi @lucalavezzo, thank you for creating the PR. Could you clarify the reason behind the change of runs? These 2 new runs are not being protected for long term usage.

jeyserma:
We briefly discussed here in the office. I believe it's better if we take new runs from ERA-F because the streamer format changed drastically. I know for Repack it doesn't matter much, but it's good to keep up to date with recent runs.

I secured these runs already, so these 2 runs can replace the older runs we used. Let me know if you agree.

germanfgv:
I removed the second DQM server `https://cmsweb-testbed.cern.ch/dqm/offline-test` as it's currently down.

germanfgv:
We had issues with DQM uploads due to the use of an outdated URL. After fixing the URL the replay finished successfully.

PR #4978, opened 2024-08-05
Remove custom siteWhiteList for Parking PDs

germanfgv:
This configuration is a remanent from 2022 changes, when we were testing the use of P5 resources only with Parking PDs. We can safely remove this limitation now.

See:
https://github.com/dmwm/T0/commit/29aa3b0d7072d804a962e5d2b7a54094769e18f8


germanfgv:
test syntax please

PR #4992, opened 2024-09-14
First proposal to dual agents by filtering streams

LinaresToine:
This PR attempts to add support to run 2 agents at the same time by adding two features:

* Define a list of streams to ignore and ignore them before the data is injected by the agent.
* Define a list of expected streams and a boolean parameter that configures an agent to process streams outside of that list or not

### DIFF
--- a/etc/ReplayOfflineConfiguration.py
+++ b/etc/ReplayOfflineConfiguration.py
@@ -28,6 +28,7 @@ from T0.RunConfig.Tier0Config import setEnableUniqueWorkflowName
 from T0.RunConfig.Tier0Config import addSiteConfig
 from T0.RunConfig.Tier0Config import setStorageSite
 from T0.RunConfig.Tier0Config import setExtraStreamDatasetMap
+from T0.RunConfig.Tier0Config import setSecondaryAgentStreams
 
 # Create the Tier0 configuration object
 tier0Config = createTier0Config()
@@ -43,6 +44,12 @@ setInjectRuns(tier0Config, [382686, 382726])
 
 # Use this in order to limit the number of lumisections to process
 #setInjectLimit(tier0Config, 10)
+
+# Define streams to ignore. These wont be injected
+setSecondaryAgentStreams(tier0Config, ["ScoutingPF", "L1ScoutingSelection", 
+                                       "L1Scouting", "Express", "Calibration",
+                                       "ExpressAlignment"
+                                       ])
 
 # Settings up sites
 processingSite = "T2_CH_CERN"
--- a/etc/ReplayOfflineConfiguration.py
+++ b/etc/ReplayOfflineConfiguration.py
@@ -28,7 +28,7 @@ from T0.RunConfig.Tier0Config import setEnableUniqueWorkflowName
 from T0.RunConfig.Tier0Config import addSiteConfig
 from T0.RunConfig.Tier0Config import setStorageSite
 from T0.RunConfig.Tier0Config import setExtraStreamDatasetMap
-from T0.RunConfig.Tier0Config import setMultipleAgentStreams
+from T0.RunConfig.Tier0Config import setHelperAgentStreams
 
 # Create the Tier0 configuration object
 tier0Config = createTier0Config()
@@ -46,9 +46,9 @@ setInjectRuns(tier0Config, [382686, 382726])
 #setInjectLimit(tier0Config, 10)
 
 # Define streams to ignore. These wont be injected
-setMultipleAgentStreams(tier0Config, {'SecondAgent': ["ScoutingPF", "L1ScoutingSelection", "L1Scouting", "Express"],
-                                      'ThirdAgent' : ["Calibration", "HLTMonitor"]
-                                      })
+setHelperAgentStreams(tier0Config, {'SecondAgent': ["ScoutingPF", "L1ScoutingSelection", "L1Scouting", "Express"],
+                                    'ThirdAgent' : ["Calibration", "HLTMonitor"]
+                                    })
 
 # Settings up sites
 processingSite = "T2_CH_CERN"
--- a/etc/ReplayOfflineConfiguration.py
+++ b/etc/ReplayOfflineConfiguration.py
@@ -28,7 +28,7 @@ from T0.RunConfig.Tier0Config import setEnableUniqueWorkflowName
 from T0.RunConfig.Tier0Config import addSiteConfig
 from T0.RunConfig.Tier0Config import setStorageSite
 from T0.RunConfig.Tier0Config import setExtraStreamDatasetMap
-from T0.RunConfig.Tier0Config import setSecondaryAgentStreams
+from T0.RunConfig.Tier0Config import setMultipleAgentStreams
 
 # Create the Tier0 configuration object
 tier0Config = createTier0Config()
@@ -46,10 +46,9 @@ setInjectRuns(tier0Config, [382686, 382726])
 #setInjectLimit(tier0Config, 10)
 
 # Define streams to ignore. These wont be injected
-setSecondaryAgentStreams(tier0Config, ["ScoutingPF", "L1ScoutingSelection", 
-                                       "L1Scouting", "Express", "Calibration",
-                                       "ExpressAlignment"
-                                       ])
+setMultipleAgentStreams(tier0Config, {'SecondAgent': ["ScoutingPF", "L1ScoutingSelection", "L1Scouting", "Express"],
+                                      'ThirdAgent' : ["Calibration", "HLTMonitor"]
+                                      })
 
 # Settings up sites
 processingSite = "T2_CH_CERN"
--- a/etc/ReplayOfflineConfiguration.py
+++ b/etc/ReplayOfflineConfiguration.py
@@ -45,9 +45,9 @@ setInjectRuns(tier0Config, [382686, 382726])
 # Use this in order to limit the number of lumisections to process
 #setInjectLimit(tier0Config, 10)
 
-# Define streams to ignore. These wont be injected
-setHelperAgentStreams(tier0Config, {'SecondAgent': ["ScoutingPF", "L1ScoutingSelection", "L1Scouting", "Express"],
-                                    'ThirdAgent' : ["Calibration", "HLTMonitor"]
+# Define streams to ignore. These wont be injected by the MainAgent
+setHelperAgentStreams(tier0Config, {'SecondAgent': [],
+                                    'ThirdAgent' : []
                                     })
 
 # Settings up sites
@@ -116,7 +116,7 @@ setPromptCalibrationConfig(tier0Config,
 
 # Defaults for CMSSW version
 defaultCMSSWVersion = {
-    'default': "CMSSW_14_0_15"
+    'default': "CMSSW_14_0_16"
 }
 
 # Configure ScramArch
--- a/src/python/T0/StorageManager/StorageManagerAPI.py
+++ b/src/python/T0/StorageManager/StorageManagerAPI.py
@@ -19,7 +19,9 @@ def injectNewData(dbInterfaceStorageManager,
                   minRun = None,
                   maxRun = None,
                   injectRun = None,
-                  injectLimit = None):
+                  injectLimit = None,
+                  isMainAgent = True,
+                  secondaryAgentStreams = []):
     """
     _injectNewData_
 
@@ -72,7 +74,15 @@ def injectNewData(dbInterfaceStorageManager,
                                     transaction = False)
 
     # remove already processed files
-    newData[:] = [newFile for newFile in newData if newFile['p5_id'] not in knownStreamers]
+    # Filtering streams for Main and Secondary agents before injecting
+    # Note that every agent is Main by default
+    # Note that the secondaryAgentStreams works as a specifyStreams if the agent is secondary
+    if isMainAgent:
+        logging.info("This is a Main Agent. The following streams will not be injected: {}".format(secondaryAgentStreams))
+        newData[:] = [newFile for newFile in newData if newFile['p5_id'] not in knownStreamers and newFile['stream'] not in secondaryAgentStreams]
+    else:
+        logging.info("This is a Secondary Agent. Only the following streams will be injected: {}".format(secondaryAgentStreams))
+        newData[:] = [newFile for newFile in newData if newFile['p5_id'] not in knownStreamers and newFile['stream'] in secondaryAgentStreams]
 
     logging.debug("StoragemanagerAPI: found %d new files", len(newData))
 
@@ -87,6 +97,7 @@ def injectNewData(dbInterfaceStorageManager,
 
         if run not in newRunStreams:
             newRunStreams[run] = set()
+        
         if stream not in newRunStreams[run]:
             newRunStreams[run].add(stream)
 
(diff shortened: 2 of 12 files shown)


LinaresToine:
Hello @jeyserma @germanfgv 

I have updated this PR so that we have two new functionalities. They are similar to ignoreStream() and specifyStreams(), except that they intend to take effect before injecting data so the handshake is not messed up when using two agents simultaneously.

The functionalities are:
* config.Global.IgnoreStreams = [] 
   * List with all streams to ignore. 
   * Any unknown/new stream will be processed
* config.Global.InjectStreams = [] 
   * List with all streams to inject
   * Should be the same as the IgnoreStreams list. 
   * Will not process unknown/new streams

Of course, the idea is to have one agent use IgnoreStreams and the other use InjectStreams

What do you guys think?

germanfgv:
We already have `specifyStreams` and `config.Global.SpecifiedStreamNames`:
https://github.com/dmwm/T0/blob/49c87fe94d1ec3baea98532884f39d0d7539aa41/src/python/T0/RunConfig/Tier0Config.py#L893-L902

Why do we need new functions and global parameters?

jeyserma:
Apart from the technical details, as of now we have the ignoreStream() and specifyStreams() functions, that act on WF level. Do we want to keep this functionality? Or do we want to merge these functions and act on the SM level?

LinaresToine:
I understand we have these functionalities already. However, we currently inject all data from the run regardless of if we use them or not. We want to inject only the streams we intend to process. I thought of making it a global variable because it should be parallel to the injection of runs variables.

However, I understand we care about the handshake because it plays a role with the repacking of injected streams. If one agent checks on ignored streams, the other agent will not repack them (this may be an imprecise comment, please correct). We could also fix the issue and use two agents by modifying the injectMinRun and MaxRun functions so that Repacking can happen regardless of the handshake. Or add another set of injectRuns functions so that it functions as in replay, where the agent process runs that are available, not runs that are just being transferred to us. This way, an agent can take care of doing handshake for all streams, while allowing the other agent to repack its corresponding streams

drkovalskyi:
Guys, please ensure that we don't lose data simply because neither agent is configured to process a given stream. At the same time, we also need to avoid both agents processing the same stream. I suspect we need to designate a master agent. Whatever solution we implement, it must handle things correctly by default. We can’t rely solely on the configuration being correct every time - our config files are far too complex for that.

LinaresToine:
I have updated the PR to take into account the feedback provided. Now, the only thing to add in the configuration file is the list of streams to ignore: 
* https://github.com/dmwm/T0/blob/e20fb63cc2728221129ee41ce483ad35805db622/etc/ReplayOfflineConfiguration.py#L49

By default, every agent is a Master agent, and the streams in the IgnoreStreams list will be ignored. This is seen in 
* https://github.com/dmwm/T0/blob/e20fb63cc2728221129ee41ce483ad35805db622/src/python/T0Component/Tier0Feeder/Tier0FeederPoller.py#L46

In order to use a Slave agent simultaneously, the following line needs to be explicit in the local file `$config/config.py`
```
config.Tier0Feeder.isMasterAgent = False
```

To complete this task, the deployment scripts need to be updated to include this new parameter in the config.py file

src/python/T0/StorageManager/StorageManagerAPI.py
```diff
@@ -87,6 +89,15 @@ def injectNewData(dbInterfaceStorageManager,
...
         if run not in newRunStreams:
             newRunStreams[run] = set()
+        
+        # If is a master agent, and the stream is included in the ignore list, dont inject it
+        # If is a slave agent, and the stream is injected by master (not in masterIgnoreStreams), it should ignore it
+        if isMasterAgent and stream in masterIgnoreStreams:
+            continue
+        
+        if not isMasterAgent and stream not in masterIgnoreStreams:
+            continue
```
germanfgv, line 99: This is preventing the agent to add the new run/stream combination, but we are still adding the streamer files.

germanfgv: We need to remove files that belong to the ignored streams from `newData` before this `for` loop starts.

germanfgv (commented):
We need to directly remove the undesired files directly from the result of the SM query. The current logic prevents the creation of the workflows, but still injects the files into the database.

LinaresToine:
It turns out the problem was not in the StorageManager code. The injectNewData function is updating the STREAM table as expected. The issue is with RunConfigAPI.py:

* Tier0FeederPoller uses [RunConfigAPI.configureRun](https://github.com/dmwm/T0/blob/master/src/python/T0Component/Tier0Feeder/Tier0FeederPoller.py#L213), which also updates the STREAM table according to the hlt mapping: 
   * https://github.com/dmwm/T0/blob/master/src/python/T0/RunConfig/RunConfigAPI.py#L130

I have filtered out the streams there also just for testing purposes, although I believe an better workaround is possible.

LinaresToine:
@jeyserma @germanfgv Please take a look at the latest changes. 
* We are now filtering out the undesired streams right after they have been queried from [GetHLTConfig.py](https://github.com/dmwm/T0/blob/7d4d4a85c923e6fdd460a170dad3c8eeaccce8fe/src/python/T0/WMBS/Oracle/RunConfig/GetHLTConfig.py)
* We are also filtering out run-stream pairs queried from [FindNewRunStreams.py](https://github.com/dmwm/T0/blob/7d4d4a85c923e6fdd460a170dad3c8eeaccce8fe/src/python/T0/WMBS/Oracle/Tier0Feeder/FindNewRunStreams.py)

I also noticed that the new [setExtraStreamDatasetMap](https://github.com/dmwm/T0/blob/master/etc/ReplayOfflineConfiguration.py#L166) function should contain streams intended to be processed by the main agent. Otherwise, they will be included in the database:
* https://github.com/dmwm/T0/blob/master/src/python/T0Component/Tier0Feeder/Tier0FeederPoller.py#L203

I added a function that takes care of this in case there is an operational mistake in the configuration:
* [resolveExtraStreamSecondaryStreamConflict](https://github.com/dmwm/T0/blob/71137bd4d8b48f12677b9bbdabbf903f436b8bd7/src/python/T0Component/Tier0Feeder/Tier0FeederPoller.py#L352)

I have tested this patch in replay and all tables of the database include the expected streams only. I believe we can update the production nodes with this patch for the cosmics runs test

germanfgv:
@LinaresToine 

> * We are now filtering out the undesired streams right after they have been queried from [GetHLTConfig.py](https://github.com/dmwm/T0/blob/7d4d4a85c923e6fdd460a170dad3c8eeaccce8fe/src/python/T0/WMBS/Oracle/RunConfig/GetHLTConfig.py)
> * We are also filtering out run-stream pairs queried from [FindNewRunStreams.py](https://github.com/dmwm/T0/blob/7d4d4a85c923e6fdd460a170dad3c8eeaccce8fe/src/python/T0/WMBS/Oracle/Tier0Feeder/FindNewRunStreams.py)

I don't like that we are filtering twice. Also, we should not modify the query from HLT. In one hand we have the database query, and on the other hand we have our own operational logic. We are modifying our logic, not the data that we get from HLT.We can add or remove whatever we want _after_ we get the HLT menu. That way it is easier to undestand if an issue comes from HLT database of from our own configuration. Further more, that allows us to filter in one single place: after we received the mapping and we added any extra streams from L1.

jeyserma:
The test with the current patch was successful.

I agree with German's comment. On MM we suggested filtering the streams on the ```hltConfig``` dictionary, just before the ```RunConfigAPI.configureRun()``` is called. In that way, there is no need anymore to filter the streams in the HLT query and also the ```resolveExtraStreamSecondaryStreamConflict``` is not necessary anymore.

LinaresToine:
test syntax please

LinaresToine:
Hello @jeyserma @germanfgv 
I have cleaned up the PR and the latest commit proposes an alternative. Please take a look

src/python/T0/RunConfig/RunConfigAPI.py
```diff
@@ -53,6 +53,39 @@ def extractConfigParameter(configParameter, era, run):
     else:
         return configParameter
 
+def filterStreams(isMainAgent = True, secondaryAgentStreams = [], hltStreamMapping=None, streamers=None):
```
germanfgv, line 56: I'm not sure why this function is part of RunConfigAPI. It's not setting any configuration nor is it used by any of the other functions on the API.

LinaresToine: This was the big question I ran into yesterday. I decided to put it there because it allows Tier0FeederPoller and StorageManagerAPI to access it. I ran into importing problems when I tried defining it in the Tier0FeederPoller. Alternatives I thought are:

* Define it in the Tier0FeederPoller
   * Split [StorageManagerAPI.injectNewData](https://github.com/dmwm/T0/blob/master/src/python/T0/StorageManager/StorageManagerAPI.py#L15) so the Tier0FeederPoller can have access to the [newData](https://github.com/dmwm/T0/blob/master/src/python/T0/StorageManager/StorageManagerAPI.py#L68) variable and filter it. I believe this sounds like the best option, but requires heavier modifications, so I wanted your opinion on it.
* Define it in the StorageManagerAPI
   * This works, and is neat, and Tier0FeederPoller would have access to it to filter out hltConfig data. The inconsistency is that it would also be used for hlt data. Perhaps that is not a problem at all

@germanfgv

LinaresToine:
An alternative to support two agents has been proposed. This one has several benefits:
* Freedom to add additional features in the future if desired 
* Freedom to deploy two agents, or three
* Safety against operational errors:
   * If the HelperAgents are not defined in the configuration, the agent works as normal
   * If the HelperAgents are defined, the agent requires a name in order to process anything. Such name has two options:
      *  "MainAgent": Ignores all streams defined for the HelperAgents
      *  Match a key in the HelperAgents dictionary: If the name is non existent, nothing is processed
      * If no name is given nothing is processed
* Easy to enable:
```
config.Tier0Feeder.agentName = "MainAgent"
```

I'll comment on the code to provide further details and walkthrough.

Finally, this has been tested in replay and the behavior is as expected. This is ready to  be tested in production. @jeyserma @germanfgv

etc/ReplayOfflineConfiguration.py
```diff
@@ -44,6 +45,11 @@
 # Use this in order to limit the number of lumisections to process
 #setInjectLimit(tier0Config, 10)
 
+# Define streams to ignore. These wont be injected
+setHelperAgentStreams(tier0Config, {'SecondAgent': ["ScoutingPF", "L1ScoutingSelection", "L1Scouting", "Express"],
```
LinaresToine, line 49: Here we provide a dictionary whose keys are agent names (arbitrary).
If a key is 'SecondAgent' then, to process the respective streams, we need
```
config.Tier0Feeder.agentName = "SecondAgent"
```

Also, name is arbitrary. We only need the name to match whatever we write in the config.py file. If there is no match, nothing is processed.

src/python/T0Component/Tier0Feeder/MultipleAgents/HelperAgent.py
```diff
@@ -0,0 +1,46 @@
...
+    Note: if the helperName is none or unknown, returned values will be empty and no data is injected
+    """
+    def __init__(self, tier0Config, helperName = None):
+        
+        BaseAgent.__init__(self, tier0Config)
+        self.name = helperName
+        if self.name and self.name in self.helperAgentStreams:
+            self.streams = self.helperAgentStreams[self.name]
+        else:
+            self.streams = []
```
LinaresToine, line 20: If there is no match between the given name and the keys in the dictionary defined in the configuration, no streams will be processed by the helper agent.

src/python/T0Component/Tier0Feeder/MultipleAgents/MainAgent.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def findUnwantedStreams (self):
+        """
+        _findUnwantedStreams_
+
+        Returns a list of streams that will not be processed by any other agent
+        """
+        self.rejectStreams = []
+        for agent in self.helperAgentStreams.keys():
+            self.rejectStreams += self.helperAgentStreams[agent]
```
LinaresToine, line 48: The MainAgent ignores every stream of each of the HelperAgents

src/python/T0Component/Tier0Feeder/Tier0FeederPoller.py
```diff
@@ -48,6 +51,9 @@ def __init__(self, config):
                                      dbinterface = myThread.dbi)
 
         self.tier0ConfigFile = config.Tier0Feeder.tier0ConfigFile
+
+        self.agentName = getattr(config.Tier0Feeder, "agentName", None)
```
LinaresToine, line 55: We get the agentName given in 
```
$config/config.py
```

src/python/T0Component/Tier0Feeder/Tier0FeederPoller.py
```diff
@@ -154,14 +161,30 @@ def algorithm(self, parameters = None):
             #
             # replays call data discovery only once (and ignore data status)
             #
+
+            # If no helper agents are given in the configuration, 
+            # this is a main agent that will ignore no streams
+            if not tier0Config.Global.HelperAgentStreams:
```
LinaresToine, line 167: If no HelperAgents are defined, we run with a MainAgent who will ignore no streams

src/python/T0Component/Tier0Feeder/Tier0FeederPoller.py
```diff
@@ -154,14 +161,30 @@ def algorithm(self, parameters = None):
...
+                logging.info("No HelperAgent provided. This is a MainAgent. Processing all streams")
+                self.agentName = "MainAgent"
```
LinaresToine, line 169: If no HelperAgents are defined, we run with a MainAgent who will ignore no streams

src/python/T0Component/Tier0Feeder/Tier0FeederPoller.py
```diff
@@ -210,6 +234,8 @@ def algorithm(self, parameters = None):
                         continue
 
                 try:
+                    if self.agentType:
+                        hltConfig['mapping'] = self.agentType.filterHltConfigStreams(hltConfig['mapping'])
```
LinaresToine, line 238: We actually use the new filtering function for hlt mapping

src/python/T0/StorageManager/StorageManagerAPI.py
```diff
@@ -72,10 +73,17 @@ def injectNewData(dbInterfaceStorageManager,
...
+    # Filtering streams for Main and Secondary agents before injecting
+    # Note that every agent is Main by default
+    # Note that the secondaryAgentStreams works as a specifyStreams if the agent is secondary
 
+    newData[:] = [newFile for newFile in newData if newFile['p5_id'] not in knownStreamers]
     logging.debug("StoragemanagerAPI: found %d new files", len(newData))
 
+    if agentType:
+        logging.info("Filtering streamer files for %s", agentType.name)
+        newData = agentType.filterStreamerFiles(streamerFiles = newData)
```
LinaresToine, line 85: We actually use the new filtering function for streamer files

germanfgv (approved):
Looks good to me. If this has been tested in replay, I think we can move forward.

germanfgv:
Seems to me we can merge this. What do you think @LinaresToine @jeyserma ?

jeyserma:
For me, it's fine to merge.

Two questions that can be answered later as well:
- is the new mechanism compatible with our replay machines?
- do we want to deprecate the other ignoreStream functionality?

PR #5000, opened 2024-10-09
Replay configuration for CMSSW_14_0_18, using also recent cosmic data

fabiocos:
# Replay Request

**Requestor**  
ORM

**Describe the configuration**  
* Release: CMSSW_14_0_18
* Run: 382686, 382726, 386674
* GTs:
   * expressGlobalTag: 140X_dataRun3_Express_v3
   * promptrecoGlobalTag: 140X_dataRun3_Prompt_v4
* Additional changes:

**Purpose of the test**  
Test of updated cosmic tracking reconstruction and DQM sequence cleaning to address memory issues in harvesting
If successful, an update of processing version for the deployment is suggested

**T0 Operations cmsTalk thread**  
https://cms-talk.web.cern.ch/t/replay-request-for-cmssw-14-0-18/52367/2

PR #5002, opened 2024-10-10
Production configuration for CMSSW_14_0_18

fabiocos:
Based on the successful output of the replay #5000 

https://cms-talk.web.cern.ch/t/replay-request-for-cmssw-14-0-18/52367/3 

I propose the update of the production configuration to CMSSW_14_0_18
The processing version is moved forward to 2, as there is a change in the behaviour of the cosmics processing, and the DQM output is also modified.


fabiocos:
@LinaresToine I believe that also the Replay configuration should be updated, no?

PR #5009, opened 2024-10-22
Reduce replay data lifetime

germanfgv:
Most replay data is only usefull for a couple of days at most. Let's reduce data lifetime in preparation for HI.


LinaresToine:
Thank you @germanfgv. Could you please include the change in the HI replay config file?

germanfgv:
@LinaresToine done :)

PR #5014, opened 2024-10-25
Unifying HI and PP configurations for HI collisions test

germanfgv:
HI and PP configurations in a single file. To be deployed in the before unstable Pb-Pb collisions during pp era `Run2024J`.

### DIFF
--- a/etc/HIProdOfflineConfiguration.py
+++ b/etc/HIProdOfflineConfiguration.py
@@ -8,7 +8,9 @@ from __future__ import print_function
 from T0.RunConfig.Tier0Config import addDataset
 from T0.RunConfig.Tier0Config import createTier0Config
 from T0.RunConfig.Tier0Config import setAcquisitionEra
+from T0.RunConfig.Tier0Config import setEmulationAcquisitionEra
 from T0.RunConfig.Tier0Config import setDefaultScramArch
+from T0.RunConfig.Tier0Config import setScramArch
 from T0.RunConfig.Tier0Config import setBaseRequestPriority
 from T0.RunConfig.Tier0Config import setBackfill
 from T0.RunConfig.Tier0Config import setBulkDataType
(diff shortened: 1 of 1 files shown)


etc/ProdOfflineConfiguration.py
```diff
@@ -1508,14 +1559,352 @@
...
+               write_nanoaod=False,
+               write_dqm=True,
+               tape_node="T0_CH_CERN_MSS",
+               disk_node="T2_US_Vanderbilt",
+               alca_producers=["TkAlMinBias"],
+               dqm_sequences=["@common"],
+               scenario=hiScenario)
+
+DATASETS = ["HIOnlineMonitor", "HITrackerNZS"]
+HIStreams += ["DQM", "PhysicsHITrackerNZS"]
```
missirol, line 1597:
```suggestion
HIStreams += ["HIDQM", "PhysicsHITrackerNZS"]
```
Starting 2024, most of the DQM streams in the PbPb menu will be named `HIDQM*`, to distinguish them more clearly from the pp menu (https://github.com/cms-sw/cmssw/pull/43645). I was not aware that HLT also needed to communicate to T0 the names of streams, in addition to the names of the Primary Datasets.

etc/ProdOfflineConfiguration.py
```diff
@@ -1508,14 +1559,352 @@
...
+               timePerEvent=1,
+               scenario=hiScenario)
 #######################
 ### ignored streams ###
 #######################
 
 ignoreStream(tier0Config, "Error")
 ignoreStream(tier0Config, "HLTMON")
 ignoreStream(tier0Config, "EventDisplay")
+ignoreStream(tier0Config, "HIEventDisplay")
```
missirol, line 1907: `HIEventDisplay` is the name of the Primary Dataset, while the name of the corresponding stream is `HIDQMEventDisplay`.

etc/ProdOfflineConfiguration.py
germanfgv: This is not tipically the case. Our current setup require the use of stream names to divide the work into several agents. We are working on changing this so we can rely on PDs in the future.

etc/HIProdOfflineConfiguration.py
```diff
@@ -1153,11 +1166,61 @@
...
+            "TestHLTPhysics8", "TestHLTPhysics9","TestHLTPhysics10","TestHLTPhysics11",
+            "TestHLTPhysics12", "TestHLTPhysics13","TestHLTPhysics14","TestHLTPhysics15",
+            "TestHLTPhysics16", "TestHLTPhysics17","TestHLTPhysics18","TestHLTPhysics19",
+            "TestHLTPhysics20", "TestHLTPhysics21","TestHLTPhysics22","TestHLTPhysics23",
+            "TestHLTPhysics24", "TestHLTPhysics25","TestHLTPhysics26","TestHLTPhysics27",
+            "TestHLTPhysics28", "TestHLTPhysics29","TestHLTPhysics30","TestHLTPhysics31",
+            "TestHLTPhysics32", "TestHLTPhysics33","TestHLTPhysics34","TestHLTPhysics35",
+            "TestHLTPhysics36", "TestHLTPhysics37","TestHLTPhysics38","TestHLTPhysics39"]
+
+STREAMS_DAQ_TFTEST = DATASETS_DAQ_TFTEST
```
missirol, line 1185: Fwiw, this (`STREAMS_DAQ_TFTEST`) and L1211 (`STREAMS_DAQ_TFTEST_ppRef`) are entirely accurate, because in this the number of streams is half of the number of PDs.

The streams are
```
TestHLTPhysics[0-19]
TestHLTPhysicsA[0-4]
TestHLTPhysicsB[0-4]
TestHLTPhysicsC[0-4]
TestHLTPhysicsD[0-4]
```

germanfgv: That's ok from our side. As long as all streams are included in the list, it works.

PR #5015, opened 2024-10-28
Fix HIZeroBias streams

jeyserma:
Fix HIZeroBias


LinaresToine:
I dont think its worth deploying this in production at this point, would you agree @jeyserma @germanfgv ?

germanfgv:
@LinaresToine seems good to me. @jeyserma I'd suggest you close the issue, unless you want to add something.

PR #5016, opened 2024-10-30
Final HI Configuration File

LinaresToine:
HIPhysicsRawPrime is planned to be splitted among streams that we reconstruct during all HIRun2024A and those we plan to stop prompt for HIRun2024B


LinaresToine:
The second commit includes the agent splitting I have noted so far: `PhysicsHIPhysics[15-59]` for the `SecondAgent`

etc/HIProdOfflineConfiguration.py
```diff
@@ -1805,25 +1832,24 @@
...
-              "PhysicsHIPhysicsRawPrime28", "PhysicsHIPhysicsRawPrime29", "PhysicsHIPhysicsRawPrime30", "PhysicsHIPhysicsRawPrime31", 
-              "PhysicsHIPhysicsRawPrime32", "PhysicsHIPhysicsRawPrime33", "PhysicsHIPhysicsRawPrime34", "PhysicsHIPhysicsRawPrime35", 
-              "PhysicsHIPhysicsRawPrime36", "PhysicsHIPhysicsRawPrime37", "PhysicsHIPhysicsRawPrime38", "PhysicsHIPhysicsRawPrime39", 
-              "PhysicsHIPhysicsRawPrime40", "PhysicsHIPhysicsRawPrime41", "PhysicsHIPhysicsRawPrime42", "PhysicsHIPhysicsRawPrime43", 
-              "PhysicsHIPhysicsRawPrime44", "PhysicsHIPhysicsRawPrime45", "PhysicsHIPhysicsRawPrime46", "PhysicsHIPhysicsRawPrime47", 
-              "PhysicsHIPhysicsRawPrime48", "PhysicsHIPhysicsRawPrime49", "PhysicsHIPhysicsRawPrime50", "PhysicsHIPhysicsRawPrime51", 
-              "PhysicsHIPhysicsRawPrime52", "PhysicsHIPhysicsRawPrime53", "PhysicsHIPhysicsRawPrime54", "PhysicsHIPhysicsRawPrime55", 
-              "PhysicsHIPhysicsRawPrime56", "PhysicsHIPhysicsRawPrime57", "PhysicsHIPhysicsRawPrime58", "PhysicsHIPhysicsRawPrime59"]
+
+StreamsHIRawPrime_reco_to_stop = ["PhysicsHIPhysicsRawPrime15",
```
jeyserma, line 1836: Instead of using "reco_to_stop" as an identifier for the streams, can you mark them as StreamsHIRawPrime_secondAgent (or something like this)? "stop" reco can confuse people when reviewing/inspecting the config file.
(and also if LHC performs badly we might be able to do full reco)

jeyserma:
The configuration looks OK to me. 

The time per event in this configuration are:
- RawPrime: 3 s/ev (measured last year 2 s/ev)
- HIForward: 1s /ev (measured last year 0.2 s/ev). We could go to 0.5 s/ev but let's keep 1 for now and monitor it.

LinaresToine:
Thanks for the comments @jeyserma , I have made the changes. I believe the only thing remaining is the new CMSSW_14_1_X release, which is yet to be available and tested. Also, @mandrenguyen mentioned new skims as well. Would you please clarify what skims to add and which PDs require these new skims?

mandrenguyen:
Hi @LinaresToine
The skims and PD mapping are specified below, which will be included in a patch release to be built this afternoon.
https://github.com/cms-sw/cmssw/blob/20891cbc2f8f2de0e9f391af58585015ac4f0f3f/Configuration/Skimming/python/autoSkim.py#L43C1-L49C112

etc/HIProdOfflineConfiguration.py
```diff
@@ -101,7 +101,7 @@
 # maxRunPreviousConfig = 999999 # Last run before era change 08/09/23
 # Defaults for CMSSW version
 defaultCMSSWVersion = {
-    'default': "CMSSW_14_1_4_patch1",
+    'default': "CMSSW_14_1_4_patch4",
```
mmusich, line 104:
```suggestion
    'default': "CMSSW_14_1_4_patch5",
```

given the outcome of https://github.com/dmwm/T0/pull/5017

etc/HIReplayOfflineConfiguration.py
```diff
@@ -111,25 +107,25 @@
 
 # Defaults for CMSSW version
 defaultCMSSWVersion = {
-    'default': "CMSSW_14_1_3"
+    'default': "CMSSW_14_1_4_patch4"
```
mmusich, line 110:
```suggestion
    'default': "CMSSW_14_1_4_patch5"
```
given the outcome of https://github.com/dmwm/T0/pull/5017.

LinaresToine:
@germanfgv This PR has been updated with the era `HIRun2024A` and `CMSSW_14_1_4_patch5`. The second agent streams have also been defined.

Please cross check one last time before deploying

germanfgv:
> Please cross check one last time before deploying

I made a small change to link cmssw version to era change, as not to use patch5 in PromptReco for era J runs. Other than than it all seems good. I'll deploy the changes.

mmusich:
I think we still miss the `UPCMonopole` skim on `HIForwardX` PDs (as tested at https://github.com/dmwm/T0/pull/5017).

PR #5021, opened 2024-11-12
Potential era change to HiRun2024B

LinaresToine:
Potential era change to `HIRun2024B` in order to prompt reconstruct all HIPhysicsRawPrime datasets, which were not being reconstructed because of potential lack of disk space to handle all the data. With a week of HI gone by, we re-evaluate the situation and perhaps we can handle all of it. To be followed up on Thursday during the meeting.

@germanfgv @jeyserma


davidlange6:
could this be instead a processing version change? (perhaps there are other reasons to change era..)

germanfgv:
> could this be instead a processing version change? (perhaps there are other reasons to change era..)

We need an Era change to make it very clear what needs to be Reco-ed later, as it divides the RAW datasets.

PR #5024, opened 2024-11-19
Support dbs block metadata in t0datasvc

LinaresToine:
This PR works in parallel to https://github.com/dmwm/t0wmadatasvc/pull/50 to provide the support requested in https://its.cern.ch/jira/browse/CMSTZDEV-812?filter=-1


germanfgv:
@LinaresToine what is the purpose of this feature? Also, do you have an estimation of the data volumes this would  upload?

LinaresToine:
@germanfgv the purpose of this feature is to have available the true completion of an era. Regarding the volumes, seems like a lot with the current setup. I just thought of an alternative to only get the count of blocks not uploaded, rather than the blocks themselves.

PR #5025, opened 2024-11-19
remove alca producers in HIRawPrime

LinaresToine:
Remive alca producers from Raw Prime PDs to reduce output volume

### DIFF
--- a/etc/HIProdOfflineConfiguration.py
+++ b/etc/HIProdOfflineConfiguration.py
@@ -135,15 +135,15 @@ hiRawPrimeScenario = "ppEra_Run3_pp_on_PbPb_approxSiStripClusters_2024"
 
 # Defaults for processing version
 alcarawProcVersion = {
-    'default': 1
+    'default': 2
 }
 
 defaultProcVersionReco = {
-    'default': 1
+    'default': 2
 }
 
 expressProcVersion = {
-    'default': 1
+    'default': 2
 }
 
 # Defaults for GlobalTag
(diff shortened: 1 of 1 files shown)


etc/HIProdOfflineConfiguration.py
```diff
@@ -1851,6 +1840,37 @@
...
+    addDataset(tier0Config, dataset,
+               do_reco=True,
+               raw_to_disk=False,
+               aod_to_disk=False,
+               timePerEvent=3,
+               write_nanoaod=False,
+               write_dqm=False,
+               tape_node="T0_CH_CERN_MSS",
+               disk_node="T2_US_Vanderbilt",
+               alca_producers=["EcalUncalZElectron", "EcalUncalWElectron", "MuAlCalIsolatedMu",
```
mmusich, line 1866: Tbh, instead of lowering the rates of all the producers by running them on less PDs (including the ones that are lightweight and low rate to start with, e.g. the dimuon resonance based once), can't alca turn off or prescale the heavy ones (e.g. `SiStripCalMinBias`) via changing the AlCaReco TriggerBits (a condition in GT)?

jeyserma: They considered it but the fastest way to reduce the ALCAs is now to simply remove them from the Tier-0 config.

mmusich: > the fastest way to reduce the ALCAs is now to simply remove them from the Tier-0 config.

How can that be faster than uploading a new payload? That takes 30s

jeyserma: If we remove them via the Tier-0 configuration we will not produce them starting from run 388462 (we currently delayed prompt for this reason). If we do it with the trigger bits it will only be effective from newer runs.

We'll deploy it tomorrow, so if there are still concerns, we can discuss it tomorrow morning.

@consuegs FYI

mmusich: >  If we do it with the trigger bits it will only be effective from newer runs.

that's incorrect.
Current FCSR from https://cmsweb.cern.ch/t0wmadatasvc/prod/firstconditionsaferun is 388462.
The alcareco trigger bit tag [AlCaRecoHLTpaths8e29_5e33_v14_prompt](https://cms-conddb.cern.ch/cmsDbBrowser/list/Prod/tags/AlCaRecoHLTpaths8e29_5e33_v14_prompt) has prompt synchronization, meeting that it will accept a new IOV with since 388462. The two approaches are equivalent in terms of rapidity of deployment, with the DB-based one fulfilling the ultimate goal of having such a mechanism and not penalizing other producer that do not contribute to data volume. Please do consider this.

jeyserma: Thanks for the clarification Marco, I wasn't aware it works via the IOV. Let's discuss the details on MM.

jeyserma:
After discussion on MM, we agree to go ahead with this Tier-0 configuration change. I reviewed the modifications and they look good to me.

mmusich:
> After discussion on MM, we agree to go ahead with this Tier-0 configuration change. I reviewed the modifications and they look good to me.

I agree, for the next time (year?) using https://github.com/cms-sw/cmssw/pull/46742 might help to do the same DB-wise.

PR #5026, opened 2024-11-22
Era by run

LinaresToine:
Proposal to change era after a specified run value.

### DIFF
--- a/etc/OXYProdOfflineConfiguration.py
+++ b/etc/OXYProdOfflineConfiguration.py
@@ -60,7 +60,53 @@ addSiteConfig(tier0Config, "T0_CH_CERN_Disk",
 #  Data type
 #  Processing site (where jobs run)
 #  PhEDEx locations
-setAcquisitionEra(tier0Config, "Run2025C")
+
+### Set Acquisition Era
+"""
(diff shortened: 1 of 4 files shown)


LinaresToine:
Currently running a replay with this patch and runs 380128, 382686, 386925.

Era `Tier0_REPLAY_2024A` is set to end at run 382686, while era `Tier0_REPLAY_2024B` at run 9999999. The workflows have been created properly.

@germanfgv @jeyserma what do you guys think of this solution?

LinaresToine:
Hello @jeyserma this patch has correctly changed era provided a specific limit run. Could you please have a look at the code?

LinaresToine:
rebuild

jeyserma:
Thanks Antonio. Apart from 1 comment, the PR looks good to me. Can you also add in the prod/replay some comments on how it should be used? I.e., you can only append new ERAs, and the one with the maximum run will be taken into account; the rests gets ignored but kept for bookkeeping reasons. 

I also think it will affect ERA extraction on the nodeSummary page?

LinaresToine:
Hello @jeyserma I have updated the changes and tested them successfully. Please review and merge when ready

PR #5035, opened 2025-01-23
Change CMS HLT DB account

jeyserma:
We have a new ```CMS_HLT_T0_R``` account to read the necessary info from the HLT database:

```
cms_hlt_gdr.u_streams
cms_hlt_gdr.u_datasets
cms_hlt_gdr.u_paths
cms_hlt_gdr.u_confversions
cms_hlt_gdr.u_pathid2strdst
cms_hlt_gdr.u_streamids
cms_hlt_gdr.u_datasetids
cms_hlt_gdr.u_pathids
cms_hlt_gdr.u_pathid2conf
cms_hlt_gdr.u_conf2strdst
```

Previously this account also reads the ```CMS_OMS``` and ```CMS_RUNINFO``` tables, but with this PR it is changed to the SMDB account which also has access to these tables.

We also need to update the SECRETS file in our agents with the new HLT credentials.

### DIFF
--- a/src/python/T0/StorageManager/StorageManagerAPI.py
+++ b/src/python/T0/StorageManager/StorageManagerAPI.py
@@ -45,10 +45,6 @@ def injectNewData(dbInterfaceStorageManager,
                                           logger = logging,
                                           dbinterface = dbInterfaceStorageManager)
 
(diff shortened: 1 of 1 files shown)


LinaresToine:
Thank you @jeyserma for this patch. Secrets file has been updated in all machines and the current production node has been patched.

PR #5041, opened 2025-03-06
Repack raw skims

LinaresToine:
This PR introduces the new feature of RAW skims in T0 produced by T0.

A raw skim is applied to an original primary dataset, for example `ParkingDoubleMuonLowMass0`, in order to get a `RAW` file that does not include all triggers in the respective streamer file. A raw skim with name `ReserveDMu` applied to `ParkingDoubleMuonLowMass0` would be processed as an independent dataset with name `ParkingDoubleMuonLowMass0-ReserveDMu`, so that the `parentDataset` `ParkingDoubleMuonLowMass0` also produces the complete `RAW` file. 

The T0 code uses dataset configurations in three critical parts:
1.  [configureRun](https://github.com/dmwm/T0/blob/f35bee6845f83d65c03a6741dfc5398aa967db5b/src/python/T0/RunConfig/RunConfigAPI.py#L133-L168)
   * Here, the datasets being looped are those in the hlt configuration, so we have no access to the defined datasets in [ReplayOfflineConfiguration.py](https://github.com/dmwm/T0/blob/f35bee6845f83d65c03a6741dfc5398aa967db5b/etc/ReplayOfflineConfiguration.py#L585-L586). So until now, explicitly defining the datasets serves no use. In this part of the code we populate the database with stream-dataset-trigger mapping.
2. [configureRunStream](https://github.com/dmwm/T0/blob/f35bee6845f83d65c03a6741dfc5398aa967db5b/src/python/T0/RunConfig/RunConfigAPI.py#L464-L570)
   *  Here, the datasets looped are those found in the database. Since we populated the database before, then the new `raw skim` datasets are included. In this part we create the output modules and configure other important details like storage sites and more. For this step we have to make sure that:
      * The output modules are added for the new `raw skim` datasets
      * Each output module is added only once
      * Each output module is included in all the steps of the loop, so none miss out on important configuration parameters.
   * The previous restrictions, while obvious, make it difficult to have the `raw skim` datasets properly defined if we only have a `raw skim` attribute in their respective `parentDatasets`. For this reason, we need to explicitly define the `raw skim` datasets in the configurtion file, and  have access to the `raw skim` name in the existing `raw skim` datasets: [datasetConfig.MyRawSkim](https://github.com/dmwm/T0/blob/f35bee6845f83d65c03a6741dfc5398aa967db5b/src/python/T0/RunConfig/RunConfigAPI.py#L476). Also, since the name of these datasets depend on their `parentDataset`, I have included `datasetConfig.ParentDataset` as well.
3. [releasePromptReco](https://github.com/dmwm/T0/blob/f35bee6845f83d65c03a6741dfc5398aa967db5b/src/python/T0/RunConfig/RunConfigAPI.py#L857).
   * Here, everything has been done. The loop takes datasets from the database and access their configuration as defined in the Replay config file. 
   
Now, I will talk about the changes in [Tier0Config.py](https://github.com/dmwm/T0/blob/f35bee6845f83d65c03a6741dfc5398aa967db5b/src/python/T0/RunConfig/Tier0Config.py).

We need to add two types of datasets: `datasets` and `raw skim datasets`
   * For the `datasets`, they need to have `datasetConfig.RawSkims = ['rawSkim1', 'rawSkim2'...]` in order to populate the database in [configureRun](https://github.com/dmwm/T0/blob/f35bee6845f83d65c03a6741dfc5398aa967db5b/src/python/T0/RunConfig/RunConfigAPI.py#L133-L168).
   * For `raw skim datasets`, we need them to be named `parentDataset-rawSkim`, and we need them to have an attribute through which I can access the actual `raw skim`, and most importantly we need them to have the same configuration as their `parent dataset` as default.  I have decided not to mix up the function [retrieveDatasetConfig](https://github.com/dmwm/T0/blob/f35bee6845f83d65c03a6741dfc5398aa967db5b/src/python/T0/RunConfig/Tier0Config.py#L381) with more elegant logic in order to make clear the difference in the default configuration for the two types of datasets.

Finally, I explain the changes in the configuration file:
* We require a global tag to actually access the trigger paths for the skim, so this one has been included in the stream configuration.

This PR works with WMCore PR and CMSSW PR


davidlange6:
in the cmssw sense, why does this need to be handled separately from other skims?

Separately - the use case for this is what ? when there is a mistake online and two things that should be separate primary datasets get merged? (a repack stage splitting would seem more resource efficient (and corresponds to whats needed in run4)

jeyserma:
> in the cmssw sense, why does this need to be handled separately from other skims?

For repacking we don't use the ConfigBuilder but we directly create the PSet with outputModules.

> 
> Separately - the use case for this is what ? when there is a mistake online and two things that should be separate primary datasets get merged? (a repack stage splitting would seem more resource efficient (and corresponds to whats needed in run4)

The need for a RawSkim at Tier-0 traces back 1.5 years ago for the ReservedDMu dataset, which was removed from the menu to save bandwidth. We started the development end of last year and decided to make it as generic as possible. 

We don't touch on how primary datasets are created. We just add an output module to the PSet with an extra path for the skim(s), which takes the trigger bits from the GT. PPD/TSG are responsible for the right content in the GT.

PR #5046, opened 2025-03-18
add CosmicsHLTMonitor

LinaresToine:
This PR contains changes proposed in: https://cms-talk.web.cern.ch/t/new-primary-dataset-cosmichltmonitor-for-cosmics-hlt-menus/118912

### DIFF
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -526,6 +526,21 @@ for dataset in DATASETS:
                sizePerEvent=155,
                scenario=cosmicsScenario)
 
+DATASETS = ["CosmicsHLTMonitor"]
+
+for dataset in DATASETS:
+    addDataset(tier0Config, dataset,
+               do_reco=True,
+               write_reco=False,
+               write_aod=True,
+               write_miniaod=False,
+               write_nanoaod=False,
+               write_dqm=True,
+               dqm_sequences=["@HLTMon"],
+               timePerEvent=0.5,
+               sizePerEvent=155,
+               scenario=cosmicsScenario)
+
 DATASETS = ["DisplacedJet"]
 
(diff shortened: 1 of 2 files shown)


etc/ProdOfflineConfiguration.py
```diff
@@ -526,6 +526,20 @@
...
+for dataset in DATASETS:
+    addDataset(tier0Config, dataset,
+               do_reco=True,
+               write_reco=False,
+               write_aod=True,
+               write_miniaod=False,
+               write_nanoaod=False,
+               write_dqm=True,
+               timePerEvent=0.5,
+               sizePerEvent=155,
```
mmusich, line 540: can we add the DQM sequences here?
it should be the same as `HLTMonitor`:

https://github.com/dmwm/T0/blob/a65498263e78e20fc8eced63e2e86acb1c09d7c5/etc/ProdOfflineConfiguration.py#L316

LinaresToine: Thanks for pointing this out @mmusich

etc/ProdOfflineConfiguration.py
```diff
@@ -526,6 +526,21 @@
                sizePerEvent=155,
                scenario=cosmicsScenario)
 
+DATASETS = ["CosmicsHLTMonitor"]
+
+for dataset in DATASETS:
+    addDataset(tier0Config, dataset,
```
mmusich, line 532: sorry @LinaresToine it just occurred to me, but `HLTMonitor` is an expressly reconstructed PD, see

https://github.com/dmwm/T0/blob/a65498263e78e20fc8eced63e2e86acb1c09d7c5/etc/ProdOfflineConfiguration.py#L310

I think it should be the same for `CosmicsHLTMonitor`

PR #5054, opened 2025-03-31
New alca producer in SpecialMinimumBias PDs

LinaresToine:
Request to add alca producer to SpecialMinimumBias PDs: [cms-talk](https://cms-talk.web.cern.ch/t/addition-of-tkalminbias-producer-under-specialminimumbias-pds/120091/2)

### DIFF
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -1272,7 +1272,7 @@ DATASETS = ["SpecialMinimumBias0", "SpecialMinimumBias1"]
 
 for dataset in DATASETS:
     addDataset(tier0Config, dataset,
-               do_reco=False,
+               do_reco=True,
                alca_producers=["TkAlMinBias"]
                )
     
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -1268,6 +1268,14 @@ for dataset in DATASETS:
                alca_producers=["SiStripCalZeroBias", "SiStripCalMinBias", "TkAlMinBias"],
                scenario=ppScenario)
 
+DATASETS = ["SpecialMinimumBias0", "SpecialMinimumBias1"]
+
+for dataset in DATASETS:
+    addDataset(tier0Config, dataset,
+               do_reco=False,
+               alca_producers=["TkAlMinBias"]
+               )
+    
 ########################################################
 ### ZeroBias PDs                                     ###
 ########################################################
--- a/etc/ReplayOfflineConfiguration.py
+++ b/etc/ReplayOfflineConfiguration.py
@@ -1118,7 +1118,7 @@ DATASETS = ["SpecialMinimumBias0", "SpecialMinimumBias1"]
 
 for dataset in DATASETS:
     addDataset(tier0Config, dataset,
-               do_reco=False,
+               do_reco=True,
                alca_producers=["TkAlMinBias"]
                )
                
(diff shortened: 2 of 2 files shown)


etc/ProdOfflineConfiguration.py
```diff
@@ -1268,6 +1268,14 @@
                alca_producers=["SiStripCalZeroBias", "SiStripCalMinBias", "TkAlMinBias"],
                scenario=ppScenario)
 
+DATASETS = ["SpecialMinimumBias0", "SpecialMinimumBias1"]
+
+for dataset in DATASETS:
+    addDataset(tier0Config, dataset,
+               do_reco=True,
+               alca_producers=["TkAlMinBias"]
+               )
```
francescobrivio, line 1277: Which values will be used for the rest of the configurables (`write_dqm`, `tape_node`, `disk_node`, `dqm_sequences`, `scenario`...) ?

LinaresToine: The default configuration, defined in [ProdOfflineConfiguration.py line 223](https://github.com/dmwm/T0/blob/5a277586b4de3b2883c64ec61aed27f59df1b170/etc/ProdOfflineConfiguration.py#L223) unless otherwise specified.

francescobrivio: Thanks @LinaresToine, looks good then!

francescobrivio:
To summarize my understanding on this PR: 
- it is currently not possibile to test the `TkAlMinBias` alcareco on the `SpecialMinimumBias*` datasets because the datasets were never produced before, so we would need the 900 GeV collisions to test it
- **BUT** this alcareco (intended for TkAlignment for 13.6 TeV collisions) and dataset (intended for BPH analyses) are exactly needed for the 900 GeV collisions

So we are in a "catch-22" situation.

Possible solutions I see:
1. Merge this PR and have everything run in production in Tier0 trusting that this AlCaReco is one of the oldest (and very stable) and the new PD is basically "MinBias with more rate".
   - Pro: if everything goes smoothly then all the needed datasets for TkAl and BPH are produced immediately
   - Con: if something goes bad Tier0 jobs will crash for this dataset/stream
2. Only run the reco for the new dataset in Tier0 and run the AlCaReco only _a posteriori_ with a small dedicated ReReco compaign which I believe can be setup fairly quick.
    - Pro: No (or minimal) risk of crashing any Tier0 production job
    - Con: TkAl might get the alcareco needed for the alignment needed for 13.6 TeV collisions a few days late
    
Tier0 (@jeyserma @LinaresToine ) and AlCa (@perrotta @atpathak @consuegs) please let me know your preferred solution.

LinaresToine:
I think we can go for option 1 provided that we are getting two new streams: `PhysicsSpecialMinimumBias0` and `PhysicsSpecialMinimumBias1`

Worst case we fail the jobs, which is better than not running them at all in my opinion. Since the new stream does not map to other PDs, then failing the jobs do not imply additional loss. Besides, we can always retry the jobs with new releases.

Let me know what you think @francescobrivio @mmusich

jeyserma:
Hi @francescobrivio if the risk of crashes at Tier-0 is very low, we can go for it. In the very worst case we have paused jobs, we can hold these jobs, prepare a new patch release and resume them with this new release.

LinaresToine:
@jeyserma and I are in sync

francescobrivio:
@jeyserma @LinaresToine thanks a lot for the quick answers!
In this case I think there is no need to further discuss this.

This will become relevant for 900 GeV collisions, which are currently expected around 18-19 April, so I let you handle the merging/deployment of these changes as you better prefer: either now or later (when other changes to the config will be applied) should be good.

LinaresToine:
I will merge now. If we dont get the stream, nothing will happen, so we might as well avoid merging later than desired

PR #5063, opened 2025-04-24
Remove GlobalTagTransaction from Express specs

jeyserma:
Removal of GlobalTagTransaction in Express workflows. It was removed from CMSSW and subsequently from WMCore. We need to remove it also from Tier-0.

See discussion here: https://github.com/dmwm/WMCore/issues/12320


LinaresToine:
Thank you Jan for providing this patch

PR #5064, opened 2025-04-24
Make Data as default setupLabel

jeyserma:
On rare occasions, the setupLabel we fetch from SM does not exist in the SM database. This prevents further runs from being processed. With this PR, we set the default setupLabel to "Data" in case it does not exist in the SM DB.

More discussion: https://its.cern.ch/jira/browse/CMSTZ-1092


jeyserma:
Yes the tests were fine (forgot the return statement but now it's OK)

LinaresToine:
Ok, I will merge then. Thank you Jan

PR #5065, opened 2025-04-25
New p ds

LinaresToine:
Following the request in [cms-talk](https://cms-talk.web.cern.ch/t/new-pds-in-the-first-2025-hlt-pp-menu/124123/2), the following datasets are being added to the configuration:

* EGamma[2-3]
* ParkingHH[0-1]
* ParkingLLP[0-1]
* ParkingSingleMuon[12-15]
* ParkingAnomalyDetection
* EmittanceScan[0-5]

Also, a few requested changes in the Physics skims of ParkingLLP and JetMET PDs.

### DIFF
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -674,7 +674,7 @@ for dataset in DATASETS:
                disk_node="T2_CH_CERN",
                scenario=ppScenario)
     
-DATASETS = ["ParkingLLP"]
+DATASETS = ["ParkingLLP0", "ParkingLLP1"]
 PARKING_PDS += DATASETS
 for dataset in DATASETS:
     addDataset(tier0Config, dataset,
@@ -682,6 +682,7 @@ for dataset in DATASETS:
                write_dqm=True,
                aod_to_disk=False,
                dqm_sequences=["@common", "@jetmet"],
+               physics_skims=["EXODelayedJet", "EXODTCluster", "EXOLLPJetHCAL"],
                archival_node=None,
                tape_node="T0_CH_CERN_MSS",
                disk_node="T2_CH_CERN",
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -726,7 +726,7 @@ for dataset in DATASETS:
                dqm_sequences=["@common"],
                scenario=ppScenario)
 
-DATASETS = ["JetMET", "JetMET0", "JetMET1"]
+DATASETS = ["JetMET0", "JetMET1"]
 
 for dataset in DATASETS:
     addDataset(tier0Config, dataset,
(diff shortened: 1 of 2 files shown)


etc/ProdOfflineConfiguration.py
```diff
@@ -882,7 +903,7 @@
                physics_skims=["ZMu", "LogError", "LogErrorMonitor"],
                scenario=ppScenario)
 
-DATASETS = ["EGamma", "EGamma0", "EGamma1"]
+DATASETS = ["EGamma", "EGamma0", "EGamma1", "Egamma2", "EGamma3"]
```
missirol, line 906:
```suggestion
DATASETS = ["EGamma", "EGamma0", "EGamma1", "EGamma2", "EGamma3"]
```

etc/ProdOfflineConfiguration.py
```diff
@@ -674,14 +694,15 @@
                disk_node="T2_CH_CERN",
                scenario=ppScenario)
     
-DATASETS = ["ParkingLLP"]
+DATASETS = ["ParkingLLP0", "ParkingLLP1"]
```
missirol, line 697: FTR, removing `ParkingLLP` and `ParkingHH` (in favour of `ParkingHH{0,1}` and `ParkingLLP{0,1}`) is okay with HLT, in the sense that today no 2025 menus contain the `ParkingLLP` and `ParkingHH` PDs.

I do not necessarily know whether or not it is okay to remove them in general from this config, that's to be determined by T0 and the other relevant experts.

PS. The same argument for removal applies to `EGamma`, `JetMET`, `DisplacedJet`, etc, etc.

etc/ProdOfflineConfiguration.py
```diff
@@ -646,6 +657,15 @@
...
                scenario=ppScenario)
 
+DATASETS = ["EmittanceScan0", "EmittanceScan1", "EmittanceScan2", 
+            "EmittanceScan3", "EmittanceScan4", "EmittanceScan5"]
+
+for dataset in DATASETS:
+    addDataset(tier0Config, dataset,
+               do_reco=True,
+               aod_to_disk=False,
+               scenario=ppScenario)
```
missirol, line 667: In the CMSTalk post, I mentioned that these are "prompt" PDs (because I do not know any better, they are not for "Parking" per se). It's a new set of PDs, for a pretty narrow use case of BRIL, and they are supposed to be small in the grand scheme of things.

I wonder if they should be configured similarly to PDs like `JetMET*` or `SpecialZeroBias*`, which seem to have a 2nd copy of RAW on T1 tape. As for AlCaRecos that BRIL may need on these EmittanceScan PDs, I do not actually know. Maybe this is something that AlCa experts could easily clarify with BRIL.

The treatment of these PDs offline is a bit beyond my expertise, but if you think TSG (and not PPD or T0) should follow this up with BRIL, please let me know.

etc/ReplayOfflineConfiguration.py
```diff
@@ -587,13 +587,26 @@
                do_reco=True,
                scenario=ppScenario)
 
-DATASETS = ["ParkingSingleMuon","ParkingSingleMuon0","ParkingSingleMuon1","ParkingSingleMuon2"]
+DATASETS = ["ParkingSingleMuon1","ParkingSingleMuon2","ParkingSingleMuon3",
```
missirol, line 590: This does not seem to have `ParkingSingleMuon0` (while the HLT menu does have it).

missirol (commented):
Thanks for the PR, Antonio. Please consider improving the title.

I left a couple of comments / questions.

etc/ProdOfflineConfiguration.py
LinaresToine: Thanks @missirol, this PDs will take the default configuration, so the current T1 tape site is FNAL

LinaresToine:
Thank you @missirol for all your comments and apologies for the multiple concentration mistakes. I have updated the changes

etc/ProdOfflineConfiguration.py
missirol: Hi Antonio, okay, thanks for the clarification.

> As for AlCaRecos that BRIL may need on these EmittanceScan PDs, I do not actually know. Maybe this is something that AlCa experts could easily clarify with BRIL.

On this, I let AlCa experts follow up with BRIL, if needed (tagging @perrotta).

PR #5066, opened 2025-04-28
Era change to Run2025B

LinaresToine:
With the new HLT Menu for Run2025, which includes new datasets as requested in https://cms-talk.web.cern.ch/t/new-pds-in-the-first-2025-hlt-pp-menu/124123/3 , and with the first 13 TeV collisions, an era change to Run2025B is proposed for the end of this week.

### DIFF
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -157,6 +157,9 @@ promptrecoGlobalTag = {
     #'acqEra': {'Run2023E': "132X_dataRun3_Prompt_v2"},
     #'maxRun': {maxRunPreviousConfig: "132X_dataRun3_Prompt_v2"}
 }
+repackGlobalTag = {
+    'default': "150X_dataRun3_Prompt_v1"
+}
 
 # Mandatory for CondDBv2
 globalTagConnect = "frontier://PromptProd/CMS_CONDITIONS"
@@ -203,6 +206,7 @@ addRepackConfig(tier0Config, "Default",
                 maxLatency=1 * 3600,
                 blockCloseDelay=24 * 3600,
                 maxMemory=2000,
+                global_tag=repackGlobalTag,
                 versionOverride=repackVersionOverride)
 
 addRepackConfig(tier0Config, "ScoutingPF",
@@ -640,6 +644,7 @@ for dataset in DATASETS:
                tape_node="T0_CH_CERN_MSS",
                disk_node="T2_CH_CERN",
                scenario=ppScenario)
+    
 
 DATASETS = ["ParkingDoubleMuonLowMass1","ParkingDoubleMuonLowMass2",
             "ParkingDoubleMuonLowMass3","ParkingDoubleMuonLowMass4","ParkingDoubleMuonLowMass5",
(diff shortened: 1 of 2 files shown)


etc/ProdOfflineConfiguration.py
```diff
@@ -640,6 +644,15 @@
...
+RAWSKIM_DATASETS = ["ParkingDoubleMuonLowMass0-ReserveDMu"]
+for rawSkimDataset in RAWSKIM_DATASETS:
+    addDataset(tier0Config, rawSkimDataset,
+               do_reco=True,
+               write_dqm=True,
+               archival_node=None,
+               tape_node="T1_US_FNAL_MSS",
+               scenario=ppScenario)
 
 DATASETS = ["ParkingDoubleMuonLowMass1","ParkingDoubleMuonLowMass2",
```
jeyserma, line 657: All the ParkingDoubleMuonLowMass PDs should have the RawSkim

jeyserma: Looks good. The corresponding streams also need to be added to the SECOND_AGENT_STREAMS list: ParkingDoubleMuonLowMass[0-3]

missirol:
Hi @jeyserma

>   * DisplacedJet --> PD does not exist anymore (please remove)
> 
>   * JetHT, MET, EGamma, SingleMuon, Muon, ParkingSingleMuon, DoubleMuon @missirol
> 
>   * HLTPhysics[0-19] @missirol only HLTPhysics is used as far as I can see
> 
>   * ParkingDoubleElectronLowMass* --> all can be removed? @missirol
> 
>   * ScoutingMonitor @patinkaew please confirm
> 
>   * TestHLTPhysics (DAQ transfer tests of last year)

I confirm that these PDs (incl. `ScoutingMonitor`) are not included in any current or upcoming HLT menus, so (from the TSG point of view) it is okay to remove them from the config.

For `TestHLTPhysics*`, I guess it could be needed again in the future (in preparation for HI, or similar), but I do not know for sure. Maybe you could keep them, if re-adding them later is non-negligible work (but again, they could be removed, because not in any current trigger menus).

etc/ProdOfflineConfiguration.py
jeyserma: Also reco can be switched OFF for these reserved parking PDs

srimanob: Hi @jeyserma @LinaresToine 

To confirm, this should be for the real situation test, not permanent. Physics coordinators agree to discuss internally with PAGs first, to understand the situation, sorting the priority, and come back to O&C. This is already agreed in XEB last week and also this week.

Actually, I understand from the joint meeting that, we will test with replay. Do I misunderstand that?

Thx.

jeyserma: Hi @srimanob 

The idea is to test it in a replay once the first collision runs arrive, and then put it directly in production once the replay is validated. Of course, this assumes that we can have this skim in production, which I thought was the case after various discussions. If you say we still need to wait for approval from management, we will still test it with a replay but not put it in production. After GL from management, it can be enabled for the next ERA 2024C at 1200 bunches.

Issue #5069, opened 2025-05-05
Possible bug: use of hltScoutingEra_Run3_2024 instead of 2025


silviodonato:
**Impact of the bug**
Likely none, as 2024 seems identical to https://github.com/cms-sw/cmssw/blob/master/Configuration/DataProcessing/python/RecoTLR.py#L97-L100

**Describe the bug**
@mtosi noticed that in https://github.com/dmwm/T0/blob/master/etc/ProdOfflineConfiguration.py#L128
we are using `hltScoutingEra_Run3_2024` instead of `hltScoutingEra_Run3_2025`.
Note that [Configuration/DataProcessing/python/Impl/hltScoutingEra_Run3_2025.py](https://github.com/cms-sw/cmssw/pull/47536/files#diff-db97b3c6af12547c6bbe1e3960c375a387176b3edb478d307f8cc16e9e57100d) was added by @srimanob with [#47536](https://github.com/cms-sw/cmssw/pull/47536), so most likely we should indeed use  `hltScoutingEra_Run3_2025`,
similarly to what was done by @LinaresToine with [#5050](https://github.com/dmwm/T0/pull/5050) for pp

cc @patinkaew

patinkaew:
Hi all,

I think there shouldn't be any impact because ```Run3_2025``` is based on ```Run3_2024``` era modifier [1] and there is no scouting-related modifier added for 2025 as far as I'm aware of. However, to avoid confusion, it might be good to change to 2025 scenario.

Best,
Patin

[1] https://github.com/cms-sw/cmssw/blob/master/Configuration/Eras/python/Era_Run3_2025_cff.py#L10

silviodonato:
This change was done in https://github.com/dmwm/T0/pull/5072
and then reverted (by mistake ? with https://github.com/dmwm/T0/pull/5071)

LinaresToine:
Thanks @silviodonato and apologies for this. The configuration file was updated.

PR #5071, opened 2025-05-14
New PD AlCaHcalIsoTrk

LinaresToine:
Request of addition of new PD: https://cms-talk.web.cern.ch/t/new-primary-dataset-for-alcahcalisotrk-hlt-menu/124762

Questions:
* Is the name `AlCaHcalIsoTrk` or `HcalCalIsoTrkFromAlCaRaw`?
* Any alca producers? Or other parameter?

@saswatinandan

### DIFF
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -101,7 +101,7 @@ setPromptCalibrationConfig(tier0Config,
 # maxRunPreviousConfig = 999999 # Last run before era change 08/09/23
 # Defaults for CMSSW version
 defaultCMSSWVersion = {
-    'default': "CMSSW_15_0_5",
+    'default': "CMSSW_15_0_6",
     #'acqEra': {'Run2024F': "CMSSW_14_0_11"},
     #'maxRun': {maxRunPreviousConfig: "CMSSW_13_2_2"}
 }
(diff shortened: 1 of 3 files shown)


mmusich:
Hi @LinaresToine 

> Is the name AlCaHcalIsoTrk or HcalCalIsoTrkFromAlCaRaw?

the name of the new PD is `AlCaHcalIsoTrk` as you correctly have:

> Any alca producers? Or other parameter?

the alca producer to be added is `HcalCalIsoTrkFromAlCaRaw`.
As for the other parameters you need at least to use `AlCaHcalIsoTrk_Run3` as processing scenario, instead of `ppScenario`.
For the remaining ones I would propose to use the same as for the other AlCaRaw dataset:

```python
do_reco=False,
raw_to_disk=True,
disk_node="T2_CH_CERN",
```

mmusich:
@LinaresToine one important piece of information, is that in order to test and deploy this PR you will need to move the cmssw release to 15_0_6 (that still doesn't exist). 
The alca producer and scenario were introduced at https://github.com/cms-sw/cmssw/pull/47975 which still is not in any release.

@JHiltbrand FYI

JHiltbrand:
Thank you @mmusich for the ping and providing necessary information !

LinaresToine:
Thanks a lot @mmusich for clarifying all the important details, I have updated the PR. 

I will now deploy a replay. Perhaps we cant really test the new PD since we don't have the corresponding stream as far as I know.

mmusich:
> Perhaps we cant really test the new PD since we don't have the corresponding stream as far as I know.

yes, the stream and PD are not yet produced, so there's no data yet to test the PR.

etc/ProdOfflineConfiguration.py
```diff
@@ -125,7 +125,8 @@
 alcaPPSScenario = "AlCaPPS_Run3"
 hiTestppScenario = "ppEra_Run3_pp_on_PbPb_2023"
 hiRawPrimeScenario = "ppEra_Run3_pp_on_PbPb_approxSiStripClusters_2023"
-hltScoutingScenario = "hltScoutingEra_Run3_2025"
```
silviodonato, line 128: Hello,
at a first look this seems a mistake (was it done on purpose?) 
Is a revert of the change requested in https://github.com/dmwm/T0/issues/5069 and applied in https://github.com/dmwm/T0/pull/5072

PR #5088, opened 2025-06-17
Add support to enable/disable repacking of emulation runs

jeyserma:
Up to now, any emulation run (test runs from SM) is repacked by default. This is not always desired as it consumes CPU and storage resources. This PR implements a flag that enables or disables the repacking of emulation runs. By default, emulation runs are not repacked.


LinaresToine:
Looks good @jeyserma thank you. Merging now

PR #5089, opened 2025-06-17
Config file for buggy replay

LinaresToine:
@Viphava280444 please use this configuration file to run a replay in vocms047.

No need to touch the deployment scripts, as they are ready for use.

Keep a close eye on it, I am expecting several different crashes. When you see the paused jobs, please document them in this PR.

### DIFF
--- a/etc/ReplayOfflineConfiguration.py
+++ b/etc/ReplayOfflineConfiguration.py
@@ -146,6 +146,8 @@ hiTestppScenario = "ppEra_Run3_pp_on_PbPb_2023"
 hiRawPrimeScenario = "ppEra_Run3_pp_on_PbPb_approxSiStripClusters_2023"
 hltScoutingScenario = "hltScoutingEra_Run3_2025"
 AlCaHcalIsoTrkScenario = "AlCaHcalIsoTrk_Run3"
+OXYScenario = "ppEra_Run3_2025_OXY"
+
 
 # Procesing version number replays
 # Taking Replay processing ID from the last 8 digits of the DeploymentID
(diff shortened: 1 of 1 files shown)


Viphava280444:
From the logs, there are 3,891 paused jobs on VOCMS047. I found that these paused jobs are in two groups, Prompt Reconstruction and Express.

In the Prompt Reconstruction group, these datasets
- ParkingDoubleMuonLowMass0
- ParkingDoubleMuonLowMass1
- ParkingDoubleMuonLowMass2
- ParkingDoubleMuonLowMass3
are paused and you can check the logs from this path: /eos/user/c/cmst0/public/PausedJobs/OhmTraining/PromptReco/SCRAMScriptFailure

Other Prompt Reconstruction jobs have a different error. They failed to open files, and you can check the logs from this path:
/eos/user/c/cmst0/public/PausedJobs/OhmTraining/PromptReco/FileOpenError

The Express jobs are also paused for the same reason — they cannot open their files and you can check the logs from this path: /eos/user/c/cmst0/public/PausedJobs/OhmTraining/Express/FileOpenError

Other Express jobs have a different error — they have exceeded the maxPSS and you can check the logs from this path:
/eos/user/c/cmst0/public/PausedJobs/OhmTraining/Express/maxPSS

LinaresToine:
Thanks a lot Ohm. It looks good so far.

One comment
* For the report we want to include the whole error message for people to see the whole context.

When dealing with real paused jobs, feel free to communicate them on cmstalk asking for experts to have a look. Provide them the tarballs as you have done here.

The next step is to decide if we resume the paused jobs or if we fail them. For now, lets fail them. You can do this with the command:

```
fail_workflow Express # fails all express paused jobs
fail_job <id> # fails single job
fail_workflow P # fails all paused jobs with a "P" in the workflow name
```

**This command is to be handled with care and after agreement with ORM**

PR #5094, opened 2025-06-25
Configuration for special runs

jeyserma:
PR to finalize the configuration for the special runs.

See cms-talk discussion for inputs: https://cms-talk.web.cern.ch/t/new-primary-dataset-for-po-oo-nene-run/126173/11


jeyserma:
Configuration converged and tested with dry runs. Merging now

PR #5104, opened 2025-07-08
Era Run2025D

LinaresToine:
pp collisions are set to begin by the end of this week. None the less, the VdM runs scheduled for tomorrow Wednesday July 8 require the era change for tomorrow morning.


LinaresToine:
@jeyserma we only have question to answer (I believe):

* How to split the streams into the next two agents

PR #5109, opened 2025-07-17
T1 sites site config to avoid sub sites

LinaresToine:
My understanding of @stlammel comment in [cmstalk](https://cms-talk.web.cern.ch/t/failing-promptreco-jobs-at-t1-es-pic-nojobreport/125179/11) is that T0 needs to explicitely specify the site config for T1 sites so that an empty value for `subsites` is propagated.

I kindly request your opinion on this Stephan. @jeyserma any concerns with this?

### DIFF
--- a/etc/ProdOfflineConfiguration.py
+++ b/etc/ProdOfflineConfiguration.py
@@ -53,6 +53,36 @@ addSiteConfig(tier0Config, "T0_CH_CERN_Disk",
                 siteLocalRucioConfig="/cvmfs/cms.cern.ch/SITECONF/T0_CH_CERN/storage.json",
                 )
 
+addSiteConfig(tier0Config, "T0_CH_CERN_Disk",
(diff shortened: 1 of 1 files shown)


LinaresToine:
My question then would be why would I configure a subsite if I dont want to use it?

stlammel:
The first arguments in overrideCatalog specify site and subsite you are at.
(THat's how the CMSSW override was implemented.)
 - Stephan

stlammel:
You would have to detect the site/subsite when the job starts and modify the overrideCatalog accordingly.
That's why i think option #2 of just setting SITECONF to T0_CH_CERN is easier.
 - Stephan

stlammel:
Yes, that should be a good place for

SITECONFIG_PATH=/cvmfs/cms.cern.ch/SITECONF/T0_CH_CERN;export SITECONFIG_PATH

 - Stephan

stlammel:
No site/subsite, you just make jobs believe they run at T0_CH_CERN.   - Stephan

PR #5079, opened 2025-05-31
Replace Oracle SEQUENCE constraints used by GENERATE auto-increment

amaltaro:
With the SQL database schema isolation and streamline, proposed in https://github.com/dmwm/WMCore/pull/12341, we need to stop using Oracle SEQUENCE constraints within the Tier0 CRUD SQL statements, as columns are now defined with the auto-increment `GENERATED BY DEFAULT AS IDENTITY` clause.

When executing a replay to test all this SQL schema separation, we need to change the T0 agent deployment to consider:
* new schema coming from https://github.com/amaltaro/wmcoredb
* same new schema above, but temporarily copied under https://github.com/dmwm/WMCore/pull/12341 as well (to be able to run Jenkins unit tests)
* in addition, WMCore codebase changes provided in the same PR above https://github.com/dmwm/WMCore/pull/12341
* lastly, the changes provided in this PR

@LinaresToine the only difference from what we have already been doing is this patch. Can you please apply it to that replay test that you started last week? Or just start a new deployment and ensure that the agent is patched with this PR. Thank you in advance!

PR #5112, opened 2025-07-25
Automate event scenario

LinaresToine:
With the merging of #5079 now the scenarios are no longer controlled by T0 in [Create.py](https://github.com/dmwm/T0/blob/master/src/python/T0/WMBS/Oracle/Create.py#L895)

We must therefore adapt to this change and automate the addition of scenarios according to what we actually use.

FYI @jeyserma


LinaresToine:
@jeyserma @Viphava280444 I removed the unnecessary changes and tested again. The event scenario table is updated without issues for all scenarios.

Please review and merge when ready

PR #5123, opened 2025-08-20
cache 5000 jobs

LinaresToine:
At T0 we want to load only 5000 jobs instead of 100000. The reason for this is to force the agent to prioritize older jobs over new jobs.

Question to consider:
* Would this imply that old prompt reco jobs are submitted before new Express jobs?

### DIFF
--- a/etc/Tier0Config.py
+++ b/etc/Tier0Config.py
@@ -20,6 +20,7 @@ config.JobSubmitter.LsfPluginResourceReq = "select[type==SLC5_64] rusage[pool=10
 config.JobSubmitter.LsfPluginJobGroup = "/groups/tier0/wmagent_testing"
 config.JobSubmitter.LsfPluginBatchOutput = "None"
 config.JobSubmitter.pollInterval = 30
+config.JobSubmitter.maxJobsToCache = 5000
 
 config.RetryManager.section_("PauseAlgo")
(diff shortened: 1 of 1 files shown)


LinaresToine:
An alternative is trying to hack directly the priority of AlcaSkim jobs by making this request of type`Skim` instead of type `Processing`

LinaresToine:
Thanks @jeyserma for clarifying. We use the query in: https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/WMBS/MySQL/Jobs/ListForSubmitter.py#L43

Where jobs are fetched by priority, where Express and Repack are higher, and then by age. I guess the bottleneck here would be the case for which Express + Repack jobs > 5000.

Looking into the database, in a huge run we got 22000+ jobs for only Express. Check the following query in vocms0314:
```
select count(*) from wmbs_job
join wmbs_jobgroup on wmbs_jobgroup.id = wmbs_job.jobgroup
join wmbs_subscription on wmbs_subscription.id = wmbs_jobgroup.subscription
join wmbs_workflow on wmbs_workflow.id = wmbs_subscription.workflow
where wmbs_workflow.name like '%395924%'
  and wmbs_workflow.name like '%Express%';
```

@jeyserma any thoughts on this?

PR #5223, opened 2026-04-28
Config for 2.4 TeV run

jeyserma:
Configuration updates for 2.4 TeV run:

Added Special2p4TeVZeroBias[0-19], parking
Changed SpecialHLTPhysics[0,31] to parking
PromptReco output goes to CERN disk

Request: https://cms-talk.web.cern.ch/t/new-pds-for-2-4-tev-run/144501


etc/ProdOfflineConfiguration.py
```diff
@@ -1182,17 +1198,46 @@
...
+               archival_node=None,
                scenario=ppScenario)
 
+DATASETS = ["Special2p4TeVZeroBias0", "Special2p4TeVZeroBias1",
+            "Special2p4TeVZeroBias2", "Special2p4TeVZeroBias3", "Special2p4TeVZeroBias4",
+            "Special2p4TeVZeroBias5", "Special2p4TeVZeroBias6", "Special2p4TeVZeroBias7",
+            "Special2p4TeVZeroBias8", "Special2p4TeVZeroBias9", "Special2p4TeVZeroBias10",
+            "Special2p4TeVZeroBias11", "Special2p4TeVZeroBias12", "Special2p4TeVZeroBias13",
+            "Special2p4TeVZeroBias14", "Special2p4TeVZeroBias15", "Special2p4TeVZeroBias16",
+            "Special2p4TeVZeroBias17", "Special2p4TeVZeroBias18", "Special2p4TeVZeroBias19"]
```
Viphava280444: Should we also add Special2p4TeVZeroBias20? In cmstalk the first thread is 0-19 but when they show us the full list. It's Special2p4TeVZeroBias[0-20]

etc/ProdOfflineConfiguration.py
```diff
@@ -1167,7 +1167,23 @@
...
                scenario=ppScenario)
 
-DATASETS = ["SpecialHLTPhysics", "SpecialHLTPhysics0", "SpecialHLTPhysics1",
+DATASETS = ["SpecialHLTPhysics"]
+
+for dataset in DATASETS:
+    addDataset(tier0Config, dataset,
+               do_reco=True,
+               raw_to_disk=True,
+               write_reco=False,
```
Viphava280444, line 1176: Why do we set SpecialHLTPhysics to 2 copies? Shouldn't we set it to one copy? Or is it because this dataset was already created previously, so we cannot change it to one copy?

etc/ProdOfflineConfiguration.py
jeyserma: Added extra PDs

etc/ProdOfflineConfiguration.py
jeyserma: In fact, we didn't receive data for this PD, but they will also not use it for the 2.4TeV runs, so it should be good.

PR #5231, opened 2026-05-19
write HIEmptyBX events at T2_CH_CERN

flodamas:
Since the HIEmptyBX dataset is mainly used for commissioning, we would like to store the events at CERN's Tier-2 on disk directly for easier access and usage.
The rate of the triggers feeding this dataset will be carefully adjusted to tens of Hz and the prescales set to zero when needed.

Thank you!
Florian for HIN and HLT

FYI @mandrenguyen , @icali , cms-trigger-coordinator@cern.ch

### DIFF
--- a/etc/HIProdOfflineConfiguration.py
+++ b/etc/HIProdOfflineConfiguration.py
@@ -1606,10 +1606,10 @@ for dataset in DATASETS:
                do_reco=True,
                write_nanoaod=False,
                write_dqm=True,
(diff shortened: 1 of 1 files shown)


mandrenguyen:
@flodamas Can you provide an estimate of the add'l disk space that will be consumed as a result of this change?

flodamas:
Hi @mandrenguyen , not more than 20 TB.

